PR #40, opened 2015-03-02
A little bit of jp translations

d-dd:


riipah:
Thanks

d-dd:
Yes  please do. I used Resx Resource Manager stand alone version and I don't think it has capability to add. Thanks.

PR #72, opened 2015-07-16
add jp strings

d-dd:
VocaDbWeb/Resources/Views/ActivityEntry/IndexStrings.ja.resx 
 VocaDbWeb/Resources/Views/Song/RankingsStrings.ja.resx 

new files


riipah:
huh, that's a lot

PR #206, opened 2016-09-12
Change to github link on help/about page

bomjacob:
Change to github link.
Small (and semi unnecessary) change, but I still thought I'd help out.


riipah:
Thanks

PR #342, opened 2017-12-17
Update to EntryMergeStrings.resx

Move-Forward39:
Is it okay for me to adjust or correct some grammar?
If so...


riipah:
Thanks!

PR #369, opened 2018-04-04
Proofreading

Move-Forward39:
I hope it's okay I did more proofreading. I removed the Delicious Cake project link since it wasn't active anymore.
I commented out the vocaloidotaku.net reference, since that's not active either, but the location is retained in case we'd like to put something else there.

### DIFF
--- a/VocaDbWeb/Views/Help/Index.cshtml
+++ b/VocaDbWeb/Views/Help/Index.cshtml
@@ -33,7 +33,7 @@
 		</p>
 		<p>
 			We specialize in Vocaloid and UTAU content, both Japanese and overseas, so the site includes multiple features designed with that in mind.
-			For that reason, the site supports embedding media, such as PVs, from NicoNicoDouga, Youtube and piapro (among other sites), that are popular among the community.
+			For that reason, the site supports embedding media, such as PVs from NicoNicoDouga, Youtube and piapro (among other sites), that are popular among the community.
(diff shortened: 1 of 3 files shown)


Move-Forward39:
The information is retained, but some things were adjusted/added for clarification.

VocaDbWeb/Views/Help/Index.cshtml
```diff
@@ -131,7 +134,7 @@
 		<br />
 		<h4 id="contactus">Contact Us</h4>
 		support[at]vocadb.net<br />
-		<a href="http://www.vocaloidotaku.net/index.php?/topic/25685-vocadb-vocaloid-database" title="">Discussion thread at VocaloidOtaku</a><br />
+		//<a href="http://www.vocaloidotaku.net/index.php?/topic/25685-vocadb-vocaloid-database" title="">Discussion thread at VocaloidOtaku</a><br />
```
riipah: Sorry, but comments like these don't work in HTML. You'll have to wrap them in HTML comment tags, see https://html.com/tags/comment-tag/

About this particular line, I think it could be just removed too, it can always be restored from source control if necessary.

Move-Forward39: Ah, thanks! I tried the double slashes and saw the line become "grayed-out" so I just assumed it worked.
I've deleted some comments and changed the others!

riipah:
Thanks!

PR #371, opened 2018-04-06
Adding Mikupa.ru banner

Move-Forward39:
I'm not sure if I did everything I need to for the banner, but here's what I did.

Also, I wasn't sure if I should keep it jpg or make it png, so I had deleted and reuploaded it a couple of times.


riipah:
Looks good. I will test tomorrow.

PR #381, opened 2018-04-18
Expanded gitignore

Euphoric:
I expanded .gitignore from https://github.com/github/gitignore/blob/master/VisualStudio.gitignore , so it is easier to work with 3rd party tools.

Previous, project-specific ignores are at the bottom.


Euphoric:
It is true that most of those entries won't be used. But it is better to use general, widely agreed upon settings, than have everyone add their own when they want to use new tool.

Euphoric:
Do you have any reasons no accepting this pull request?
I believe there is no problem using settings that is used when you select to use Github's default .gitignore for Visual Studio when creating new repository.

riipah:
In my opinion it's not necessary, and it adds unnecessary clutter. Sorry.

PR #382, opened 2018-04-18
Japanese interface translations

Move-Forward39:
I was going to wait for DD to get back to me on one of the translations (for "Artist's Personal Description"), but I think we can merge these for now. I can add it later.

DD, if you happen to see this, feel free to let me know what you think would be best for that, and I can add it. [The discussion is here, for reference](https://vocadb.net/discussion/topics/6).

I did some rearranging in some files, so if anything is syntactically incorrect, it'd be from that. (The reason for rearranging was to better match the order of the English version.) Let me know if there's an error and I can fix it.


riipah:
Thanks! Should be fine.

Issue #366, opened 2018-03-06
Show full list of comments for events


riipah:
Currently only last 3 are shown

PR #383, opened 2018-04-25
Move Forward Branch

Move-Forward39:
I did a handful of things for this branch, but there's a few things I'd like double-checked when you have time if that's alright. The XML stuff shouldn't need to be double-checked.

- Replaced "Cancel" on button with reference to string value of "Cancel" for CommentHelpers.cshtml.
- Replaced "Merge" on button with reference to the string value of "MergeBtnText" for Merge.cshtml.
^ Would you be able to double check these to make sure I did that right?

- Added translations for "Merge" from EntryEditStrings to EntryMergeStrings (assuming the translation would be the same).
- Added what I think may be part of adding comments + the "view all comments" section for Events. (#366) I'm not sure, though, so it's commented out (at least, if I did it correctly, it should be commented out).
^ Could you double check just to make sure the whole segment is correctly commented out?

- Many instances of NicoNicoDouga -> NicoNico and language equivalents (since I believe NicoNicoDouga now goes by NicoNico.)
- Deleted Description.htm.


VocaDbWeb/Views/Event/Details.cshtml
```diff
@@ -216,6 +216,29 @@
 
 @CommentHelpers.LatestCommentsKnockout()
 
+<!--	<div id="tabs" class="js-cloak" data-bind="show, with: comments">
```
riipah: This will require a bit more code. Can you please remove this change and create another branch?

riipah:
Thanks! Other changes are good, but the comments on event page won't work like that yet. Can you remove changes to that one file?

PR #401, opened 2018-07-07
Piapro links error + YouTube links

Move-Forward39:
As brought up in this discussion: https://vocadb.net/discussion/topics/157 There's been some issues with Piapro links, and I think that's because perhaps "https" was added. I added it with the piapro-https-links branch. I also added https://youtu.be/ for YouTube.
I don't know if I did everything necessary, though.


riipah:
The fix has to be made to this project: https://github.com/riipah/piapro-dotnet, more specifically here https://github.com/riipah/piapro-dotnet/blob/master/PiaproClient/PiaproClient.cs#L164 . I can see if I have time today...

riipah:
I updated PiaproClient to support https and updated the dependency to VocaDB. https URLs are working now (not published yet).

Move-Forward39:
Ahh, I see. Cool!

PR #406, opened 2018-07-14
Twipla icon issue 403

riipah:


riipah:
Note that updating the css file by hand is actually unnecessary, it's generated from the .less file automatically by Visual Studio.

Move-Forward39:
Ah, I see. Thanks!

Move-Forward39:
If I just updated the .less file on the GitHub site, would Visual Studio still automatically update the css file (like on your end maybe)?

riipah:
Yeah, when I open the project in VS

PR #481, opened 2019-03-10
Update WebLinkMatcher.ts

Nefere256:
Few matchers for links I frequently add when editing TouhouDB.

Mixcloud and MQube are music hosting services (while Mixcloud is more used by DJs) and Otoya Page (音屋ぺーじ) is for hosting porfolios.


riipah:
Thank you!

PR #520, opened 2019-11-07
Clean up CollectionHelper

ycanardeau:
The SyncWithContent method can be simplified by using the Sync method.


riipah:
Thanks! I'll try to review this tonight.

PR #521, opened 2019-11-08
Fix garbled credit text

ycanardeau:
The credit text "みゆ" in the footer is not displayed correctly due to the wrong encoding.


riipah:
Can't really see the difference in GitHub, but I guess the file got saved with UTF-8 without BOM or something? Not sure when it happened.

riipah:
Hmm, regarding this. I tried reopening the file in VS then saving, and it's broken again. It's probably because of the charset setting UTF-8 without BOM in .editorconfig. By default VS saves with BOM. In general it's recommended to skip the BOM for compatibility, but maybe the absence of BOM causes trouble for ASP.NET.

riipah:
Ah it's probably this https://stackoverflow.com/a/22319145

riipah:
Okay, I ended up specifying utf-8-bom for .cshtml files by default, instead of changing the default file encoding, because documentation says that if you specify fileEncoding then all files MUST be in that encoding. So, saving .cshtml files with BOM should be the safest. Others can be without BOM, but I guess I could save all files with BOM too...

PR #522, opened 2019-11-09
create EventController.ViewSeriesVersion

ycanardeau:


riipah:
I'll try to finish reviewing this tomorrow. Thank you!

VocaDbWeb/Views/Event/ViewSeriesVersion.cshtml
```diff
@@ -0,0 +1,55 @@
...
+		};
+	} else {
+		ViewBag.Parents = new[] {
+			Html.ActionLink("Events", "Index"),
+		};
+	}
+
+}
+
+@if (Model.Entry.ComparableVersions.Any())
```
riipah: It's a small thing, but the braces here are different than elsewhere. I'll try to add these to .editorconfig for Visual Studio at least.

VocaDbModel/Database/Queries/EventQueries.cs
```diff
@@ -632,7 +640,7 @@ public class EventQueries : QueriesBase<IEventRepository, ReleaseEvent> {
 					};
 					session.Save(series);
 
-					var diff = new ReleaseEventSeriesDiff(ReleaseEventSeriesEditableFields.OriginalName | ReleaseEventSeriesEditableFields.Names);
```
riipah: Could you change this back?

riipah:
Thank you. It seems mostly fine. Much appreciated. Could you take a look at those small comments I posted though?

VocaDbWeb/Views/Event/ViewSeriesVersion.cshtml
ycanardeau: fixed

VocaDbModel/Database/Queries/EventQueries.cs
ycanardeau: I think this shouldn't be changed back because this creates a snapshot just like other entry types.
If I change this back, the "External links" row in the "Content" table is displayed as one column rather than two columns when comparing two revisions because the WebLinks field is not included in the serialized data.

Please take the following steps to confirm the problem.
1. Create a new event series without any external links.
2. Add some external links to the series.
3. Compare these two revisions.

ycanardeau:
Should I implement this feature for song lists as well?

riipah:
Less important for song lists, but it's not really harming.

ycanardeau:
Thank you for your review and advice.

riipah:
Thanks for adding it!

Issue #124, opened 2016-02-12
Tagging song lists


riipah:
Song lists should be tagged like other entries.

riipah:
Done and deployed, thanks to @ycanardeau

PR #525, opened 2019-11-17
Implement tagging system for song lists

ycanardeau:
#124

### DIFF
--- a/VocaDbModel/Database/Queries/UserQueries.cs
+++ b/VocaDbModel/Database/Queries/UserQueries.cs
@@ -40,6 +40,7 @@ using VocaDb.Model.Service.Paging;
 using VocaDb.Model.Service.Queries;
 using VocaDb.Model.Service.QueryableExtenders;
 using VocaDb.Model.Service.Search;
+using VocaDb.Model.Service.Search.SongSearch;
 using VocaDb.Model.Service.Search.User;
 using VocaDb.Model.Service.Security;
 using VocaDb.Model.Service.Security.StopForumSpam;
@@ -1018,21 +1019,22 @@ namespace VocaDb.Model.Database.Queries {
 
 		}
 
-		public PartialFindResult<SongListForApiContract> GetCustomSongLists(int userId, SearchTextQuery textQuery, SongListSortRule sort, PagingProperties paging, SongListOptionalFields fields) {
+		public PartialFindResult<SongListForApiContract> GetCustomSongLists(int userId, SongListQueryParams queryParams, SongListOptionalFields fields) {
 			
 			return HandleQuery(ctx => { 
 				
 				var query = ctx.Query<SongList>()
 					.WhereNotDeleted()
 					.Where(s => s.Author.Id == userId && s.FeaturedCategory == SongListFeaturedCategory.Nothing)
-					.WhereHasName(textQuery);
+					.WhereHasName(queryParams.TextQuery)
+					.WhereHasTags(queryParams.TagIds, queryParams.ChildTags);
 
-				var items = query.OrderBy(sort)
-					.Paged(paging)
+				var items = query.OrderBy(queryParams.SortRule)
+					.Paged(queryParams.Paging)
 					.Select(s => new SongListForApiContract(s, userIconFactory, entryImagePersister, fields))
 					.ToArray();
 
-				var count = paging.GetTotalCount ? query.Count() : 0;
+				var count = queryParams.Paging.GetTotalCount ? query.Count() : 0;
 
 				return new PartialFindResult<SongListForApiContract>(items, count);
 
--- a/VocaDbModel/Database/Queries/UserQueries.cs
+++ b/VocaDbModel/Database/Queries/UserQueries.cs
@@ -1098,7 +1098,7 @@ namespace VocaDb.Model.Database.Queries {
 			return GetTagSelections<ReleaseEventSeries, EventSeriesTagUsage, EventSeriesTagVote>(seriesId, userId);
 		}
 
-		public TagSelectionContract[] GetSongListTagSelection(int songListId, int userId) {
+		public TagSelectionContract[] GetSongListTagSelections(int songListId, int userId) {
 			return GetTagSelections<SongList, SongListTagUsage, SongListTagVote>(songListId, userId);
 		}
 
--- a/VocaDbModel/Database/Queries/UserQueries.cs
+++ b/VocaDbModel/Database/Queries/UserQueries.cs
@@ -1096,6 +1096,10 @@ namespace VocaDb.Model.Database.Queries {
 			return GetTagSelections<ReleaseEventSeries, EventSeriesTagUsage, EventSeriesTagVote>(seriesId, userId);
 		}
 
+		public TagSelectionContract[] GetSongListTagSelection(int songListId, int userId) {
+			return GetTagSelections<SongList, SongListTagUsage, SongListTagVote>(songListId, userId);
+		}
+
 		public TagSelectionContract[] GetSongTagSelections(int songId, int userId) {
 
 			return HandleQuery(session => {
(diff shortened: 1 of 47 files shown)


riipah:
Thanks! I'll try to review tonight.

VocaDbWeb/Scripts/Repositories/SongListRepository.ts
```diff
@@ -72,6 +78,10 @@ module vdb.repositories {
 
 		};
 
+		public getTagSuggestions = (songId: number, callback: (contract: dc.tags.TagUsageForApiContract[]) => void) => {
+			$.getJSON(this.urlMapper.mapRelative("/api/songLists/" + songId + "/tagSuggestions"), callback);
```
riipah: I didn't see this method in SongListApiController. Are you intending to add tag suggestions based on songs in the list?

VocaDbModel/Database/Queries/UserQueries.cs
```diff
@@ -1096,6 +1098,10 @@ class CachedUserStats {
 			return GetTagSelections<ReleaseEventSeries, EventSeriesTagUsage, EventSeriesTagVote>(seriesId, userId);
 		}
 
+		public TagSelectionContract[] GetSongListTagSelection(int songListId, int userId) {
```
riipah: Maybe GetSongListTagSelections

VocaDbWeb/Scripts/ViewModels/SongList/FeaturedSongListsViewModel.ts
```diff
@@ -1,40 +1,44 @@
...
+				this.categories[categoryName] = new FeaturedSongListCategoryViewModel(listRepo, resourceRepo, tagRepo, languageSelection, cultureCode, categoryName);
 			});
 
-			this.initCategory(categoryNames[0]);
-
 		}
 
 		public categories: { [index: string]: FeaturedSongListCategoryViewModel; } = {};
 
-		public initCategory = (category: string) => {
```
riipah: I couldn't figure out based on the code why this method is not needed. Can you explain?

VocaDbWeb/Views/Event/Details.cshtml
```diff
@@ -232,7 +232,8 @@
 		$("#manageTags").button({ icons: { primary: 'ui-icon-wrench' } });
 
 		var loggedUserId = @UserContext.LoggedUserId;
-		var urlMapper = new vdb.UrlMapper('@UrlMapper.HostAddress');
+		var rootPath = '/';
```
riipah: Hmm why did you hardcore rootpath to '/' instead of using the config parameter?

VocaDbWeb/Views/Event/SeriesDetails.cshtml
```diff
@@ -97,7 +97,8 @@
 		$("#viewVersions").button({ icons: { primary: 'ui-icon-clock' } });
 
 		var loggedUserId = @UserContext.LoggedUserId;
-		var urlMapper = new vdb.UrlMapper('@UrlMapper.HostAddress');
+		var rootPath = '/';
```
riipah: Here too

VocaDbWeb/Views/Shared/Partials/_SongListsFilters.cshtml
```diff
@@ -1,22 +1,43 @@
...
-			</a>
-			<a data-bind="click: function() { sort('CreateDate'); }, css: { active: sort() == 'CreateDate' }" class="btn" href="#">
-				@VocaDb.Web.Resources.Views.SongList.SongListSortRuleNames.CreateDate
-			</a>
-			<a data-bind="click: function() { sort('Name'); }, css: { active: sort() == 'Name' }" class="btn" href="#">
-				@VocaDb.Web.Resources.Views.SongList.SongListSortRuleNames.Name
-			</a>
+﻿<div class="form-horizontal well well-transparent" data-bind="visible: true" style="display: none;">
+	
+	<div class="pull-right">
```
riipah: I don't know why did you change this to dropdown instead of tabs, but I guess it works like this too.

VocaDbWeb/Views/SongList/Featured.cshtml
```diff
@@ -52,10 +52,19 @@
...
 
-		var urlMapper = new vdb.UrlMapper('@RootPath');
+		var languageSelection = '@UserContext.LanguagePreference';
+		var rootPath = '@RootPath';
+		var urlMapper = new vdb.UrlMapper(rootPath);
+		var repoFactory = new vdb.repositories.RepositoryFactory(urlMapper, @LanguagePreferenceInt, @UserContext.LoggedUserId);
+		var songListRepo = repoFactory.songListRepository();
+		var resourceRepo = repoFactory.resourceRepository();
+		var tagRepo = repoFactory.tagRepository();
+		var languageSelection = '';
```
riipah: Why are you overriding languageSelection here?

riipah:
I think TagTargetTypes should be updated, even if you're not doing tag filtering by entry type (that can be a separate task).

riipah:
TagDetailsContract should be updated. At least the number of song lists.

VocaDbModel/Domain/Tags/Tag.cs
```diff
@@ -401,6 +402,14 @@ public Tag(LocalizedString name, string categoryName = "")
 			}
 		}
 
+		public virtual ISet<SongListTagUsage> AllSongListTagUsages {
```
riipah: This should be added to AllTagUsages. Events too (I forgot them before).

riipah:
TagUsageQueries.GetActualTagUsageCounts(...) should be updated too, so that it takes into account song list tag usage counts.

riipah:
Thanks for the PR! I don't blame you for missing those few things, unfortunately there's quite a lot of places that you just need to know when adding a new tag type.

VocaDbWeb/Views/SongList/Featured.cshtml
ycanardeau: I guess it must be a mistake. Thanks.

VocaDbWeb/Views/Shared/Partials/_SongListsFilters.cshtml
ycanardeau: Yes, and I think this is more consistent with other layouts.

VocaDbWeb/Views/Event/Details.cshtml
ycanardeau: Fixed, but @UrlMapper.HostAddress should not be used because it generates a non-SSL URL (e.g. http://localhost:39390/ instead of https://localhost:44399/).

VocaDbWeb/Scripts/ViewModels/SongList/FeaturedSongListsViewModel.ts
ycanardeau: This method is not needed because the clear method gets called when the getList method succeeded. If this method is present, the loadMore method gets called twice when first loading and duplicate song lists appear.

ycanardeau:
Thank you for your review!

ycanardeau:
I updated TagDetailsContract, but I haven't changed the view. By the way, the "Show more" button for event series is linked to the event search page. Is this correct?

riipah:
> I updated TagDetailsContract, but I haven't changed the view. 

Maybe you could add the number of song lists at least? And link to featured song lists (even though it's not all).

>By the way, the "Show more" button for event series is linked to the event search page. Is this correct?

Yeah that's intentional. For now event search page is also good enough for series.

ycanardeau:
Do you mean like this? Or should I add "Top song lists" to the view? I don't think "All song lists" is an appropriate name though because it is actually not "all"...

riipah:
Should be ok, thanks!

Issue #527, opened 2019-11-20
Allow reporting users for spamming


riipah:
Split from VocaDB/community#108 

* Button on user details page, titled "report for spamming".
* Description/evidence must be provided.
* All users except "limited" can report.
* Moderators cannot be reported (to prevent abuse)
* For 5 active (not closed) reports, set user to "limited" (disable editing)
* For 10 active reports, disable user.
* Trusted users can be reported, but will not be limited or disabled automatically.

riipah:
TODO: make notes mandatory.

PR #528, opened 2019-11-20
Report users for spamming

riipah:
#527

TODO: make notes mandatory (currently silently ignores). Maybe change to ReportEntryViewModel.
TODO: unit tests.
TODO: prevent multiple open reports by the same user.

### DIFF
--- a/VocaDbModel/Database/Queries/UserQueries.cs
+++ b/VocaDbModel/Database/Queries/UserQueries.cs
@@ -538,9 +538,9 @@ namespace VocaDb.Model.Database.Queries {
 
 				if (user.GroupId <= UserGroupId.Regular && reportType == UserReportType.Spamming) {
 					var activeReportCount = ctx.Query<UserReport>()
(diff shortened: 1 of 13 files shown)


VocaDbModel/Database/Queries/UserQueries.cs
```diff
@@ -516,6 +517,52 @@ class CachedUserStats {
...
+					if (activeReportCount >= 10) {
+						log.Info("User disabled");
+						user.Active = false;
+						ctx.Update(user);
+					} else if (activeReportCount >= 5) {
+						log.Info("User set to limited");
+						user.GroupId = UserGroupId.Limited;
+						ctx.Update(user);
+					}
+				}
```
ycanardeau: I think this should be done after an entry report has been created, or you'll need 5+1 active reports to set user to limited and 10+1 to disable user.

riipah: I was thinking it wasn't that important, 5 or 6 reports. But yeah, makes more sense that way. Moved it. Thank you for reviewing!

PR #540, opened 2020-01-12
Manage JavaScript dependencies with Laravel Mix

ycanardeau:
#5


ycanardeau:
#5

ycanardeau:
Please note that you have to delete all auto-generated JavaScript files (ignored by .gitignore) in the Scripts folder before building the solution.

riipah:
I'll be quite busy this week because of Miku Expo too, so I will probably not be able to review right now. Also, I'll need to check if I need to make changes to the build script because of this (probably will).

I wanted to move building to Azure DevOps, but it might be too slow.

ycanardeau:
Have fun!

By the way, are you planning to migrate from ASP.NET MVC to ASP.NET Core MVC in the future?

VocaDbWeb/Helpers/LaravelMixHelper.cs
```diff
@@ -0,0 +1,30 @@
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Web;
+using System.Web.Optimization;
+using Newtonsoft.Json;
+
+namespace VocaDb.Web.Helpers {
+
+	public static class LaravelMixHelper {
```
riipah: Could you add comments to this class?

ycanardeau: I've added some comments to the class.

riipah:
Sorry it's taking a while. It's a lot of code, and I want to make sure we are at least moving to right direction. Also, like you said, you disabled TypeScript compilation in Visual Studio / MSBuild, which means that I will need to update the server to build using npm. Right now it's just building everything with MSBuild, which also compiles JavaScript from .ts files.

The code files such as SongEditViewModel.ts look fine, with the imports and exports. I'm still a bit unclear what's the purpose of Laravel Mix, I've only used webpack directly (at work). It seems Laravel generates the webpack config file and calls webpack?

VocaDbWeb/package.json
```diff
@@ -0,0 +1,34 @@
...
+        "typescript": "^3.7.4",
+        "vue-template-compiler": "^2.6.11"
+    },
+    "dependencies": {
+        "bootstrap-sass": "^2.3.2",
+        "jquery": "^2.2.1",
+        "jquery-ui": "^1.10.4",
+        "knockout": "^3.4.1",
+        "knockout-punches": "^0.5.1",
+        "lodash": "^4.17.3",
```
riipah: Since these JS libraries are now installed using npm (which is good), do you think we can remove them from nuget (packages.config)? JS packages used to be on nuget, but not anymore, because Microsoft is encouraging people to use npm for that.

VocaDbWeb/webpack.mix.js
```diff
@@ -0,0 +1,194 @@
...
+	], "bundles/jqxRating.js")
+
+
+	// VIEW-SPECIFIC BUNDLES
+	.scripts([
+	], "bundles/ActivityEntry/Index.js")
+
+
+	.scripts([
+	], "bundles/Album/Create.js")
```
riipah: I'm not sure I understand the purpose of these scripts commands. Does this produce a file called Create.js? What's in that file?

VocaDbWeb/package.json
ycanardeau: > Since these JS libraries are now installed using npm (which is good), do you think we can remove them from nuget (packages.config)?

Yes I think so, and I've uninstalled those NuGet packages.

VocaDbWeb/webpack.mix.js
ycanardeau: Yes, this produces an empty file called Create.js to prevent 404 errors. I didn't remove those scripts commands so that I can split App.ts into page-specific scripts later. In case it would be OK to get rid of the page-specific bundles and move all of application code to the shared bundles, we can remove them, or at least we should comment them (plus the corresponding `RenderScripts` in .cshtml files) out before merging this pull request. By the way, I use `App.ts` (and `App.js`) rather than `app.ts` so that I can easily rename the `app` variable later (e.g. to `vdb`).

ycanardeau:
> It seems Laravel generates the webpack config file and calls webpack?

To be honest, I don't know the internals of Webpack and Laravel Mix very well, but [Laravel Mix doesn't have any dependencies on Laravel](https://blog.jonstodle.com/easy-webpack-with-laravel-mix/).

VocaDbWeb/webpack.mix.js
riipah: Ah okay. Yeah that's fine. Could you add a comment explaining this?

ycanardeau: I've added comments.

riipah:
I'll set up a test website (#551) and deploy it there.

Issue #542, opened 2020-01-14
Knockout.Punches converts double curly-brace syntax input by users to a virtual element


ycanardeau:
For example, if an album description contains `{{expression}}`, it is converted to `<!--ko text:expression--><!--/ko-->` by `ko.punches.enableAll()` in [vocadb/VocaDbWeb/Views/Album/Details.cshtml](https://github.com/VocaDB/vocadb/blob/a29a6a3ee2f300d87aeaa39ebaf7d039168bb515/VocaDbWeb/Views/Album/Details.cshtml#L495).

riipah:
Right. Do you think it's a real issue, would any albums have that description? Can we just escape it? Or maybe we can skip specific elements from KO punches evaluation.

ycanardeau:
I'm not sure whether or not an album has that description, but if there is one, it causes `Uncaught SyntaxError: Unable to parse bindings.` and the page cannot be loaded. I haven't found out the appropriate solution yet, so I submitted this issue.

riipah:
Perhaps we could apply StopBinding to all description/comment fields. It should take care of this.

riipah:
StopBinding seems to be working.

ycanardeau:
With this method, you'll have to apply StopBinding not only to the Description field, but also to all fields that can be edited by users. For example, if a user inputs `{{{'<i>Hello World!</i>'}}}` in the Non-English name field of an album, it is rendered as HTML (`<script>` tags don't work for some reason). I guess this is not an expected behaviour.

ycanardeau:
Fixed in #1245.

PR #543, opened 2020-01-14
#542: ignore all markdown fields from KO binding

riipah:
Add StopBinding to all Markdown-formatted server-side rendered text.

### DIFF
--- a/VocaDbWeb/Views/Album/Details.cshtml
+++ b/VocaDbWeb/Views/Album/Details.cshtml
@@ -416,7 +416,7 @@
 							<h4 class="album-review-title">@Model.LatestReview.Title</h4>
 						}
 
-						<p>@HtmlHelpers.FormatMarkdown(Model.LatestReview.Text)</p>
+						<div>@HtmlHelpers.FormatMarkdown(Model.LatestReview.Text)</div>
 					</div>
 				</div>
 			</div>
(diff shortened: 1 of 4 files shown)


VocaDbWeb/Views/Album/Details.cshtml
```diff
@@ -416,7 +416,7 @@
 							<h4 class="album-review-title">@Model.LatestReview.Title</h4>
 						}
 
-						<p>@HtmlHelpers.FormatMarkdown(Model.LatestReview.Text)</p>
+						<div>@HtmlHelpers.FormatMarkdown(Model.LatestReview.Text)</div>
```
riipah: I noticed the hard way that <p> tags cannot be inside other <p> tags, and if they do, strange things will happen.

ycanardeau:
With this method, you'll have to apply StopBinding not only to the Description field, but also to all fields that can be edited by users. For example, if a user inputs `{{{'<i>Hello World!</i>'}}}` in the Non-English name field of an album, it is rendered as HTML (`<script>` tags don't work for some reason). I guess this is not an expected behaviour.

riipah:
Since I added it to Markdown rendering, that covers at least most of the multi-line text, not just album description. But yes, it should probably be added to all server-side rendered fields. Unfortunately I can't think of a catch-all method for this.

Issue #547, opened 2020-01-26
API for event series details


riipah:
https://github.com/VocaDB/VocaDB-App/issues/103

PR #548, opened 2020-01-27
Update ReleaseEventSeriesApiController

ycanardeau:
#547

### DIFF
--- a/VocaDbModel/DataContracts/ReleaseEvents/ReleaseEventSeriesForApiContract.cs
+++ b/VocaDbModel/DataContracts/ReleaseEvents/ReleaseEventSeriesForApiContract.cs
@@ -1,0 +1,103 @@
+using System;
+using System.Linq;
+using System.Runtime.Serialization;
+using VocaDb.Model.Domain;
+using VocaDb.Model.Domain.Globalization;
+using VocaDb.Model.Domain.Images;
+using VocaDb.Model.Domain.ReleaseEvents;
+
+namespace VocaDb.Model.DataContracts.ReleaseEvents {
+
+	public class ReleaseEventSeriesForApiContract {
+
+		public ReleaseEventSeriesForApiContract() { }
+
+		public ReleaseEventSeriesForApiContract(ReleaseEventSeries series, ContentLanguagePreference languagePreference, ReleaseEventSeriesOptionalFields fields, IEntryThumbPersister thumbPersister) {
+
+			Category = series.Category;
+			Id = series.Id;
+			Name = series.TranslatedName[languagePreference];
+			Status = series.Status;
+			Version = series.Version;
+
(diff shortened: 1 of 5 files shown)


VocaDbModel/DataContracts/ReleaseEvents/ReleaseEventSeriesForApiContract.cs
```diff
@@ -0,0 +1,103 @@
...
+		[DataMember(EmitDefaultValue = false)]
+		public string AdditionalNames { get; set; }
+
+		[DataMember]
+		public EventCategory Category { get; set; }
+
+		[DataMember(EmitDefaultValue = false)]
+		public string Description { get; set; }
+
+		[DataMember]
```
riipah: I would add [DataMember(EmitDefaultValue = false)] to all optional fields, including this one. WebLinks too, I think. It's better to return no list than empty list when the field is not requested. I know for events the EmitDefaultValue isn't there, but maybe it should.

riipah: And this field should be ReleaseEventForApiContract instead of ReleaseEventContract. The "non-API" versions are mainly for internal use.

ycanardeau: Thanks for your review!

VocaDbModel/DataContracts/ReleaseEvents/ReleaseEventSeriesForApiContract.cs
```diff
@@ -29,7 +30,7 @@ public class ReleaseEventSeriesForApiContract {
 			}
 
 			if (fields.HasFlag(ReleaseEventSeriesOptionalFields.Events)) {
-				Events = series.Events.Select(e => new ReleaseEventContract(e, languagePreference)).ToArray();
+				Events = series.Events.Select(e => new ReleaseEventForApiContract(e, languagePreference, ReleaseEventOptionalFields.None, thumbPersister)).ToArray();
```
ycanardeau: What should I pass as the third argument here? Non for-api data contracts are used in other for-api data contracts (e.g. [SongForApiContract](https://github.com/VocaDB/vocadb/blob/43754bb4b75b5de0d067c775728e8a9bc2d99dc4/VocaDbModel/DataContracts/Songs/SongForApiContract.cs#L106)).

riipah: None is fine for now. It's easier to add fields than to remove them. That AlbumContract probably was forgotten there. It's better to use the Api contracts.

riipah:
Thank you!

PR #553, opened 2020-02-06
Refactoring

ycanardeau:


riipah:
We're currently having a spammer problem and mitigating that is taking all my time. I'll review this after that.

riipah:
Looks fine, thanks again!

PR #560, opened 2020-02-12
Fix TagUsageQueries

ycanardeau:
#559 Although I'm not sure if this is the right way to fix this, it seems to work fine.

### DIFF
--- a/VocaDbModel/Domain/Tags/IEntryWithTags.cs
+++ b/VocaDbModel/Domain/Tags/IEntryWithTags.cs
@@ -1,6 +1,8 @@
-﻿namespace VocaDb.Model.Domain.Tags {
+namespace VocaDb.Model.Domain.Tags {
 
 	public interface IEntryWithTags : IEntryBase {
+
+		bool AllowNotifications { get; }
 
 		ITagManager Tags { get; }
 
--- a/VocaDbModel/Service/Queries/TagUsageQueries.cs
+++ b/VocaDbModel/Service/Queries/TagUsageQueries.cs
@@ -6,6 +6,7 @@ using VocaDb.Model.Database.Repositories;
 using VocaDb.Model.DataContracts.Tags;
 using VocaDb.Model.Domain;
 using VocaDb.Model.Domain.Security;
+using VocaDb.Model.Domain.Songs;
 using VocaDb.Model.Domain.Tags;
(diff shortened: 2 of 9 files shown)


riipah:
In general it's fine. Could you try writing a unit test for that in TagUsageQueriesTests?

VocaDbModel/Service/Queries/TagUsageQueries.cs
```diff
@@ -95,7 +96,10 @@ public class TagUsageQueries {
 					entryLinkFactory.CreateEntryLink(entry), string.Join(", ", tagNames)), user);
 
 				var addedTags = appliedTags.Except(entry.Tags.Tags).ToArray();
-				new FollowedTagNotifier().SendNotifications(ctx, entry, addedTags, new[] { user.Id }, entryLinkFactory, enumTranslations);
+
+				if (!((entry is SongList songList) && !songList.FeaturedList)) {
```
riipah: Perhaps extract this to a private method bool AllowNotifications(entry). I might also consider moving the check to IEntryWithTags, but since it's just this one place, maybe it's ok to have the type check here.

ycanardeau: Thank you for your review. Since I think your solution is more elegant than mine, I added `AllowNotifications` property to `IEntryWithTags` interface. I wanted to use default interface implementation for the property, but the target runtime doesn't support it. I wrote a unit test for this, could you please review it?

riipah: Default interface implementation would be great here, but unfortunately it's only for .NET Core 3.1, and we're not quite there yet (maybe some day).

VocaDbModel/Domain/Tags/IEntryWithTags.cs
```diff
@@ -1,7 +1,9 @@
-﻿namespace VocaDb.Model.Domain.Tags {
+namespace VocaDb.Model.Domain.Tags {
 
 	public interface IEntryWithTags : IEntryBase {
 
+		bool AllowNotifications { get; }
```
riipah: Maybe add a comment here?

riipah:
The test looks fine. I can merge this, perhaps consider adding a comment for the new property.

PR #561, opened 2020-02-13
Replace UrlMapper.HostAddress with RootPath

ycanardeau:


VocaDbWeb/Views/User/Login.cshtml
```diff
@@ -8,7 +7,8 @@
 
 @Helpers.ValidationSymmaryDiv(Res.UnableToLogin)
 
-<form action="@VocaUriBuilder.Absolute(Url.Action("Login", "User"))" method="POST">
```
ycanardeau: Is there any reason why you use form tag and `VocaUriBuilder.Absolute` instead of `Html.BeginForm` here?

riipah: Can't be 100% sure, but I think it was because originally VocaDB login form was using https, while the rest of the site was http by default, so the URL would have to be absolute. Now it could be either.

VocaDbWeb/Views/Shared/Partials/_LayoutScripts.cshtml
```diff
@@ -20,7 +20,7 @@
 
     vdb = vdb || {};
     vdb.values = vdb.values || {};
-    vdb.values.hostAddress = '@UrlMapper.HostAddress';
```
riipah: Did you review all places where this is used, so it can be relative instead of absolute?

riipah: Actually, now that I'm checking context, baseAddress already contains the relative path (RootPath), right? So if absolute path is not needed, it's best to just remove this hostAddress variable and migrate remaining usages to baseAddress

ycanardeau: Removed `hostAddress` variable from GlobalValues.ts.

Issue #275, opened 2017-04-08
Add entries for event venues


riipah:
Could be used to reuse event venue information such as name, location and web address.

Venue management should be as lightweight as possible: a new venue can be created from event page. We still probably need a page for listing venues so that they can be deleted as necessary.

Who can create/edit venue information?

riipah:
@ycanardeau I made some fixes (changed lat/long input to number, fix display in cultures that use comma).

Could you add the address field to venue edit page?

ycanardeau:
Thanks for your fixes. Do you think only the address field is enough, or should I add separate fields for country, state and street address so that we can search for venues by country?

riipah:
Separate field for country (saved as [2 letter ISO 3166 country code](https://docs.microsoft.com/en-us/dotnet/api/system.globalization.regioninfo.twoletterisoregionname?view=netframework-4.8)) sounds like a good idea. Rest of the address can be one field.

riipah:
@ycanardeau seems that Venue/Edit.ts/js is missing. Can you add that file?

ycanardeau:
Added.

riipah:
Thank you again, this is now in production.

Issue #550, opened 2020-02-02
Convert project files to PackageReference and SDK format


riipah:
In order to prepare for .NET Core conversion (#544), Model, Web and Test should be converted to use PackageReference instead of packages.config.

Both PackageReference and SDK format are required by .NET Core / Standard, but recommended even for .NET Framework.

Additionally, Model and Test can be converted to use the SDK project format. SDK format is not supported for ASP.NET projects (except Core), so Web will remain in legacy format.

https://www.michaeltaylorp3.net/migrating-to-sdk-project-format/

PR #563, opened 2020-02-27
Add entries for event venues

ycanardeau:
#275 We can now search venues by latitude and longitude by using `VenueApiController`. Example: `http://localhost:39390/api/venues?latitude=39&longitude=39&radius=39&distanceUnit=Kilometers&sortRule=Distance`. I tried to use [geography data types](https://docs.microsoft.com/en-us/sql/t-sql/spatial-geography/spatial-types-geography?view=sql-server-ver15) in SQL Server along with [NHibernate.Spatial](https://github.com/nhibernate/NHibernate.Spatial), but in vain.


VocaDb.Migrations/Migrations.cs
```diff
@@ -1,17 +1,233 @@
...
+				
+			Alter.Table(TableNames.EntryReports).AddColumn("Venue").AsInt32().Nullable().ForeignKey(TableNames.Venues, "Id").OnDelete(Rule.Cascade);
+			
+			Create.Column("Venue").OnTable(TableNames.ActivityEntries).AsInt32().Nullable()
+				.ForeignKey(TableNames.Venues, "Id").OnDelete(Rule.Cascade);
+
+			Create.Column("ArchivedVenueVersion").OnTable(TableNames.ActivityEntries).AsInt32().Nullable()
+				.ForeignKey(TableNames.ArchivedVenueVersions, "Id").OnDelete(Rule.None);
+
+			Create.Column("VenueEntry").OnTable(TableNames.AlbumReleaseEvents).AsInt32().Nullable().ForeignKey(TableNames.Venues, "Id").OnDelete(Rule.SetNull);
```
ycanardeau: I wanted to use the name `Venue`, but the name is already in use.

VocaDbModel/Domain/ReleaseEvents/ReleaseEventEditableFields.cs
```diff
@@ -38,7 +38,9 @@ public enum ReleaseEventEditableFields {
 
 		Venue			= 16384,
 
-		WebLinks		= 32768
+		VenueEntry		= 32768,
```
ycanardeau: Here too.

VocaDb.Migrations/Migrations.cs
ycanardeau: I wanted to use the name `Venue`, but the name is already in use.

VocaDbModel/Domain/ReleaseEvents/ReleaseEventEditableFields.cs
ycanardeau: Here too.

VocaDbModel/Service/QueryableExtenders/VenueQueryableExtender.cs
```diff
@@ -0,0 +1,92 @@
...
+
+				var earthRadius = GetEarthRadius(distanceUnit);
+
+				// HACK: calculate the Haversine distance.
+				// TODO: do something.
+				query = query.Where(v => earthRadius * 2 * Math.Asin(Math.Sqrt(
+					Math.Pow(Math.Sin((queryParams.Latitude.Value - Math.Abs(v.Coordinates.Latitude.Value)) * Math.PI / 180 / 2), 2)
+					+ Math.Cos(queryParams.Latitude.Value * Math.PI / 180)
+					* Math.Cos(Math.Abs(v.Coordinates.Latitude.Value) * Math.PI / 180)
+					* Math.Pow(Math.Sin((queryParams.Longitude.Value - v.Coordinates.Longitude.Value) * Math.PI / 180 / 2), 2))) < radius.Value);
```
ycanardeau: Here too.

VocaDbModel/Service/QueryableExtenders/VenueQueryableExtender.cs
```diff
@@ -0,0 +1,92 @@
...
+
+			var earthRadius = GetEarthRadius(distanceUnit);
+
+			// HACK: calculate the Haversine distance.
+			// TODO: do something.
+			return query.OrderBy(v => earthRadius * 2 * Math.Asin(Math.Sqrt(
+				Math.Pow(Math.Sin((coordinates.Latitude.Value - Math.Abs(v.Coordinates.Latitude.Value)) * Math.PI / 180 / 2), 2)
+				+ Math.Cos(coordinates.Latitude.Value * Math.PI / 180)
+				* Math.Cos(Math.Abs(v.Coordinates.Latitude.Value) * Math.PI / 180)
+				* Math.Pow(Math.Sin((coordinates.Longitude.Value - v.Coordinates.Longitude.Value) * Math.PI / 180 / 2), 2))));
```
ycanardeau: We have to do something here.

VocaDb.Migrations/Migrations.cs
riipah: I think it would be better to rename the old field to VenueName.

riipah:
The distance calculation doesn't need to be that accurate... I was thinking it's enough if the country is correct. But of course even better if you can search for nearby countries/cities too.

VocaDb.Migrations/Migrations.cs
ycanardeau: If you rename the old field to `VenueName`, then you also have to rename the serialized `Venue` field in archived versions, so that revisions can be displayed correctly.

ycanardeau: It turned out to be no problem to rename Venue to VenueName in the AlbumReleaseEvents table, so I renamed it, but still the `Venue` field in the `ReleaseEventEditableFields` enum cannot be renamed easily because of the reason I said.

VocaDbWeb/Views/Venue/Details.cshtml
```diff
@@ -0,0 +1,92 @@
...
+	@Helpers.EntryStatusMessage(Model.Status)
+}
+
+@if (Model.Deleted) {
+	@EntryDetailsHelpers.DeletedBanner(null)
+}
+
+@if (Model.Coordinates.HasValue) {
+	<iframe src="https://maps.google.com/maps?q=@(Model.Coordinates.Latitude),@(Model.Coordinates.Longitude)&output=embed" frameborder="0" scrolling="no" marginheight="0" marginwidth="0" width="600" height="450"></iframe>
+}
```
ycanardeau: I added Google Maps to the venue details page.

VocaDb.Migrations/Migrations.cs
riipah: I don't understand why. Besides, the field in ArchivedEventContract is already VenueName.

riipah:
I pulled changes from master to this branch. Could you recreate the pull request so it's easier to review with less changes? I don't know if there's another way.

VocaDb.Migrations/Migrations.cs
ycanardeau: The ChangedFields columns in the ArchivedEventVersions table are saved as comma-separated string. So if we rename `Venue` in the `ReleaseEventEditableFields` enum to `VenueName`, then we also have to update ChangedFields columns. I think we'll need to be careful with changes to existing table columns.

ycanardeau:
Thank you. I've resolved conflicts and you can now review with less changes without having to recreate the pull request.

VocaDb.Migrations/Migrations.cs
riipah: Oh you mean that. It should be possible to update that SQL UPDATE command in the migration. In SQL server, REPLACE can be used to do string replace. So something like 

```sql
UPDATE [ArchivedEventVersions] SET (ChangedFields = REPLACE(ChangedFields, 'Venue', 'VenueName')) WHERE ChangedFields LIKE '%Venue%'
```

(above is not tested)

VocaDb.Migrations/Migrations.cs
```diff
@@ -1,1805 +1,1866 @@
-using System.Data;
```
riipah: I wonder, why is the whole file changed? Line endings?

VocaDb.Migrations/Migrations.cs
ycanardeau: Sorry, I was talking about the serialized XML at first, but it turned out that I was wrong.

VocaDb.Migrations/Migrations.cs
ycanardeau: I've resolved conflicts in Migrations.cs on GitHub and I'm not sure why the whole file has changed...

ycanardeau: Fixed.

VocaDbModel/DataContracts/Venues/VenueWithEventsContract.cs
```diff
@@ -0,0 +1,34 @@
+using System.Collections.Generic;
+using System.Linq;
+using VocaDb.Model.DataContracts.ReleaseEvents;
+using VocaDb.Model.Domain.Globalization;
+using VocaDb.Model.Domain.ReleaseEvents;
+using VocaDb.Model.Domain.Venues;
+
+namespace VocaDb.Model.DataContracts.Venues {
+
+	public class VenueWithEventsContract : VenueContract {
```
riipah: How about, if you add Events (as optional field) to VenueForApiContract, then you shouldn't need this contract anymore, right? I've been moving towards using those "Api" contracts with optional fields whenever possible.

VocaDbModel/DataContracts/Venues/VenueDetailsContract.cs
```diff
@@ -0,0 +1,20 @@
+using System.Linq;
+using VocaDb.Model.DataContracts.ReleaseEvents;
+using VocaDb.Model.Domain.Globalization;
+using VocaDb.Model.Domain.Venues;
+
+namespace VocaDb.Model.DataContracts.Venues {
+
+	public class VenueDetailsContract : VenueContract {
```
riipah: This could be replaced with VenueForApiContract too, I think.

VocaDbModel/Database/Queries/VenueQueries.cs
```diff
@@ -0,0 +1,288 @@
...
+				return new Model.Service.Queries.EntryReportQueries().CreateReport(ctx, PermissionContext,
+					entryLinkFactory,
+					(song, reporter, notesTruncated) => new VenueReport(song, reportType, reporter, hostname, notesTruncated, versionNumber),
+					() => reportType != VenueReportType.Other ? enumTranslations.Translation(reportType) : null,
+					venueId, reportType, hostname, notes);
+			});
+
+		}
+		
+		private void CreateTrashedEntry(IDatabaseContext ctx, Venue venue, string notes) {
```
riipah: So you created all this already huh. I was thinking we wouldn't need this for venues, but I guess it's fine since it's already there.

VocaDbModel/Domain/OptionalGeoPoint.cs
```diff
@@ -0,0 +1,77 @@
...
+
+	public interface IOptionalGeoPoint {
+
+		double? Latitude { get; }
+
+		double? Longitude { get; }
+
+	}
+
+	public class OptionalGeoPoint : IOptionalGeoPoint {
```
riipah: Maybe add a small comment here. If Latitude/Longitude is specified, the other one needs to be specified too, but it's ok for both to be empty.

VocaDb.Migrations/Migrations.cs
riipah: Actually, there's probably a FluentMigrator command for doing that update more easily too.

VocaDbModel/Domain/ReleaseEvents/ReleaseEventEditableFields.cs
riipah: I think it would be better to change the old field to VenueName and the new field just Venue, doing the migration I mentioned.

VocaDbModel/Domain/Venues/Venue.cs
```diff
@@ -0,0 +1,143 @@
...
+			}
+		}
+
+		public virtual EntryStatus Status { get; set; } = EntryStatus.Draft;
+
+		public virtual TranslatedString TranslatedName => Names.SortNames;
+
+		public virtual int Version { get; set; }
+
+		public virtual IList<VenueWebLink> WebLinks {
```
riipah: Could've used the WebLinkManager here, but I guess it doesn't matter much.

VocaDbModel/Domain/Venues/Venue.cs
```diff
@@ -0,0 +1,143 @@
...
+using VocaDb.Model.Domain.Activityfeed;
+using VocaDb.Model.Domain.ExtLinks;
+using VocaDb.Model.Domain.Globalization;
+using VocaDb.Model.Domain.ReleaseEvents;
+using VocaDb.Model.Domain.Security;
+using VocaDb.Model.Domain.Versioning;
+
+namespace VocaDb.Model.Domain.Venues {
+
+	public class Venue : IEntryWithNames<VenueName>, IEntryWithVersions<ArchivedVenueVersion, VenueEditableFields>,
```
riipah: Could you add CreateDate here still. That's for almost all entries.

VocaDbModel/Service/QueryableExtenders/VenueQueryableExtender.cs
riipah: What does the current code do?

VocaDbModel/Service/QueryableExtenders/VenueQueryableExtender.cs
```diff
@@ -0,0 +1,92 @@
...
+				// TODO: do something.
+				query = query.Where(v => earthRadius * 2 * Math.Asin(Math.Sqrt(
+					Math.Pow(Math.Sin((queryParams.Latitude.Value - Math.Abs(v.Coordinates.Latitude.Value)) * Math.PI / 180 / 2), 2)
+					+ Math.Cos(queryParams.Latitude.Value * Math.PI / 180)
+					* Math.Cos(Math.Abs(v.Coordinates.Latitude.Value) * Math.PI / 180)
+					* Math.Pow(Math.Sin((queryParams.Longitude.Value - v.Coordinates.Longitude.Value) * Math.PI / 180 / 2), 2))) < radius.Value);
+
+			}
+
+			// HACK: do not use GeoPointQueryParams.HasValue here, otherwise NHibernate.QueryException will be thrown.
```
riipah: I think you mean OptionalGeoPoint.HasValue? But yeah, NHibernate won't understand the property call.

VocaDbModel/Service/ReleaseEventService.cs
```diff
@@ -34,6 +34,25 @@ public ReleaseEventService(ISessionFactory sessionFactory, IUserPermissionContex
 
 		}
 
+		public VenueWithEventsContract[] GetReleaseEventsByVenues() {
```
riipah: You should move this to EventQueries class instead, unless there's a specific reason adding it here. These legacy "services" are tied to NHibernate, and not testable.

VocaDbModel/Service/ReleaseEventService.cs
```diff
@@ -34,6 +34,25 @@ public ReleaseEventService(ISessionFactory sessionFactory, IUserPermissionContex
...
+			return HandleQuery(session => {
+
+				var allEvents = session.Query<ReleaseEvent>().Where(e => !e.Deleted).ToArray();
+				var venues = session.Query<Venue>().Where(e => !e.Deleted).OrderByName(LanguagePreference).ToArray();
+
+				var venueContracts = venues.Select(v =>
+					new VenueWithEventsContract(v, allEvents.Where(e => v.Equals(e.Venue)), PermissionContext.LanguagePreference));
+				var ungrouped = allEvents.Where(e => e.Venue == null).OrderBy(e => e.TranslatedName[LanguagePreference]);
+
+				return venueContracts.Concat(new[] { new VenueWithEventsContract {
```
riipah: You should be able to use .Append to add one value without creating an array.

VocaDbModel/VocaDb.Model.csproj
```diff
@@ -1,1240 +1,1265 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
```
riipah: Also this file has been completely regenerated for some reason. Can you fix it?

We should change the Model project to use SDK project format anyway (see #550), so it doesn't need to list all files. It would make changing project file less common.

VocaDb.Migrations/Migrations.cs
```diff
@@ -5,15 +5,96 @@ namespace VocaDb.Migrations {
...
+
+		public override void Up() {
+			Execute.Sql("UPDATE [ArchivedEventVersions] SET ChangedFields = REPLACE(ChangedFields, 'Venue', 'VenueName') WHERE ChangedFields LIKE '%Venue%'");
+		}
+
+		public override void Down() {
+			Execute.Sql("UPDATE [ArchivedEventVersions] SET ChangedFields = REPLACE(ChangedFields, 'VenueName', 'Venue') WHERE ChangedFields LIKE '%VenueName%'");
+		}
+
+	}
```
ycanardeau: I know I should not use `Execute.Sql`, but I could not find a way other than this.

VocaDbModel/Service/QueryableExtenders/VenueQueryableExtender.cs
ycanardeau: In order to calculate the distance between two points on a sphere, I applied the [haversine formula](https://en.wikipedia.org/wiki/Haversine_formula) in a very ugly way because NHibernate won't understand property calls as you said. I actually wanted to use geography data types and NHibernate.Spatial as I said before, but they don't work well. Please see: https://build-failed.blogspot.com/2012/03/nhibernate-spatial-part-4-querying.html

ycanardeau:
All done. Thank you for always reviewing.

VocaDbModel/DataContracts/Venues/VenueDetailsContract.cs
ycanardeau: Do you think `VenueForEditContract` should be replaced with `VenueForApiContract`, too?

VocaDbModel/DataContracts/WebLinkForApiContract.cs
```diff
@@ -11,12 +11,17 @@ public class WebLinkForApiContract : IWebLink {
 
 		public WebLinkForApiContract() {}
 
-		public WebLinkForApiContract(WebLink webLink) {
+		public WebLinkForApiContract(WebLink webLink, WebLinkOptionalFields fields = WebLinkOptionalFields.None) {
```
riipah: Maybe this could have been just a simple boolean too :)

VocaDbModel/Service/QueryableExtenders/VenueQueryableExtender.cs
riipah: If it works then it's fine to keep it like this I think? It's not that ugly I think. You could mention in the comment what you just said, and you can remove the "HACK: do something" comment, I think.

VocaDb.Migrations/Migrations.cs
riipah: Using Execute.Sql is fine. But I think you should add the "WHERE ChangedFields LIKE '%Venue%' AND NOT ChangedFields LIKE '%VenueName%'... just in case.

VocaDbModel/DataContracts/Venues/VenueDetailsContract.cs
riipah: Perhaps I would keep VenueForEditContract separate. The edit use case is quite special.

VocaDbWeb/Views/Event/Details.cshtml
```diff
@@ -162,6 +162,14 @@
 	}
 }
 
+@if (Model.Venue != null) {
```
riipah: Might consider showing the Google map here too, if there's space.

VocaDbWeb/Views/Event/Edit.cshtml
```diff
@@ -179,11 +179,19 @@
...
+				@KnockoutHelpers.LockingAutoComplete("venueAutoComplete", "venue.entry", "venue.name")
+				@Html.HiddenForKnockout(m => m.Venue, "value: ko.toJSON(venue)")
+			</div>
+
 			<div class="editor-label">
 				@Helpers.HelpLabel("Venue", "Can be either a physical location such as concert hall, or a virtual location (website).")
 			</div>
 			<div class="editor-field">
-				@Html.TextBoxFor(m => m.Venue, new { size = 40, @class = "span3", maxlength = 1000 })<br />
+				@Html.TextBoxFor(m => m.VenueName, new { size = 40, @class = "span3", maxlength = 1000 })<br />
```
riipah: So you allow entering both venue link and venue name? I think that, if venue is selected from the autocomplete, then we can hide this venue name field and automatically fill it based on the selected venue. 

The venue name field can be used when there is no entry for the venue, if it's a one-off event in that venue. This should also be explained in the description.

VocaDbWeb/Views/Event/EventsByDate.cshtml
```diff
@@ -15,6 +15,9 @@
 	<li>
 		@Html.ActionLink(Res.ViewBySeries, "EventsBySeries")
 	</li>
+	<li>
+		@Html.ActionLink(Res.ViewByVenues, "EventsByVenues")
```
riipah: Actually, I would prefer getting rid of all these old ASP.NET pages. The search page and the "Upcoming" events page should be enough. I guess the EventsByVenues page can be there now that it's added, but I think it could be removed later.

VocaDbWeb/Views/Venue/Edit.cshtml
```diff
@@ -0,0 +1,162 @@
...
+	<div class="editor-field">
+		<table>			
+			<tr>
+				<td data-bind="with: names">
+					@{ Html.RenderPartial("KnockoutPartials/_NamesEditor", new VocaDb.Web.Models.Shared.NamesEditorViewModel()); }
+					@Html.HiddenForKnockout(m => m.Names, "value: ko.toJSON($data.toContracts())")
+				</td>
+				<td style="vertical-align: top">
+					<div class="alert" data-bind="visible: duplicateName" style="display: none;">
+						Series already exists with name
```
riipah: Series -> Venue

VocaDbWeb/Views/Venue/Edit.cshtml
```diff
@@ -0,0 +1,162 @@
...
+	</div>
+	<div class="editor-field">
+		@Html.TextAreaForKnockout(m => m.Description, "initialValue, textInput: description", cols: 60, rows: 4, maxLength: 1000, cssClass: "span4")
+		
+		Live preview
+		<div data-bind="markdown: description"></div>
+	</div>
+		
+	<table><tr>
+		<td>@Res.Coordinates</td>
```
riipah: I think we should really have some sort of search box for this. Or selecting the location from map. Nobody is going to fill latitude and longitude by hand.

riipah:
Thank you for adding the feature. It looks fine for the most part now. I finished with the review.

The search box for coordinates is still quite necessary I think, but it can also be split into another issue, if you want.

Please make the change for VenueName field that I requested: if Venue is already filled, then VenueName should be hidden.

ycanardeau:
Thank you for your review. It looks like we need a project with a billing account to use the [Google Maps Platform](https://developers.google.com/maps/documentation/geocoding/start).

VocaDb.Migrations/Migrations.cs
```diff
@@ -5,15 +5,96 @@ namespace VocaDb.Migrations {
...
+	[Migration(2020_03_01_1300)]
+	public class ArchivedEventVersionsRenameVenue : Migration {
+
+		public override void Up() {
+			Execute.Sql("UPDATE [ArchivedEventVersions] SET ChangedFields = REPLACE(ChangedFields, 'Venue', 'VenueName') WHERE ChangedFields LIKE '%Venue%' AND NOT ChangedFields LIKE '%VenueName%'");
+		}
+
+		public override void Down() {
+			Execute.Sql("UPDATE [ArchivedEventVersions] SET ChangedFields = REPLACE(ChangedFields, 'VenueName', 'Venue') WHERE ChangedFields LIKE '%VenueName%'");
+		}
```
ycanardeau: If the `Down` method is executed, we may not be able to distinguish `Venue` and `VenueName` anymore, because both of them become `Venue`.

riipah: Well, you could rename Venue to VenueEntry first. But I don't think it's that important.

Issue #495, opened 2019-05-14
Bilibili flash embeds don't work


AgFlore:
Bilibili flash embeds don't load on multiple browsers of mine. For example https://vocadb.net/S/162351.
Guess there to be some problem with the old flash embed.

Is is possible to update all the old flash embeds that some entries still use into the new HTML5 embed?

P.S. I would suggest adjust the height of the Bilibili HTML5 embed to show the controls. I've tried it out and finds that the threhold of showing controls is that the height of embed must be greater than 360px, sounds like not very far from the currently using 315px.

Also, according to https://blog.asucreyau.xyz/2018/11/19/bilibili-embed-optimize/, the best height of the embed should be (width*9/16+118px), where the 118px is the total height of the top banner, the time-control and the bottom danmuku input. In the case here the banner and the danmuku input is useless, so perhaps we can mask it out in the final page with some html technique?

riipah:
I'm not doing a script for this, but there is a facility available for moderators to refresh the embed for one song at a time. Users may also do that by hand of course.

Issue #494, opened 2019-04-16
Refer to the exact episode when embedding media from bilibili


AgFlore:
One bilibili video can contain multiple episodes, each might be a PV for a separate song. 
For example, the song [虚拟少年](https://vocadb.net/S/163542) was first released as live performance on the 2017 Luotianyi Concert in Shanghai, and its video is released on Bilibili as the 14th episode of the concert record, https://www.bilibili.com/video/av12802603/?p=14.
Currently, if one submit that link as a bilibili media to vocadb, the "p=14" will be ignored, and the p1 episode of the video will be embeded, which is not desired.

If one check https://api.bilibili.com/x/player/pagelist?aid=12802603, which looks like 
>...{"cid":21707899,"page":14,"from":"vupload","part":"虚拟少年","duration":250,"vid":"","weblink":"","dimension":{"width":0,"height":0,"rotate":0}},...
he shall then find the cid of that specific episode and may embed it in the page.

riipah:
This might be difficult because the system isn't designed to include multiple parameters. We did that with SoundCloud, but the implementation is a bit complicated.

szc126:
Just rediscovered this old comment about a song being hidden behind piapro's "historical revisions" that allow you to reuse the same id.
https://vocadb.net/S/7319/discussion

https://piapro.jp/content/wzbannuehci39y5c
>（前のバージョンと言うとこから、普通のカラオケ、本家がDLできます）

Issue #372, opened 2018-04-06
Replace Bilibili flash embeds with HTML-based embed code


riipah:
Example:
```<iframe src="//player.bilibili.com/player.html?aid=21140802&cid=34714040&page=1" scrolling="no" border="0" frameborder="no" framespacing="0" allowfullscreen="true"> </iframe>```

The HTML embeds require a new "cid" parameter in addition to video "aid". This cid parameter is returned for API responses, so VocaDB should save it. I'm not sure what is the purpose of that parameter, but it varies by PV.

riipah:
A problem with the new player is that the playback controls won't appear unless the player is big enough. The embedded player on VocaDB is below that size. I think it's rather silly, because you can't for example seek at all. I wonder if there's an option to force the controls to appear always?

riipah:
The new player is added on VocaDB, but the missing controls issue is not resolved yet. Also, the new player will only be available for new PVs, because it requires the cid parameter.

AgFlore:
> The HTML embeds require a new "cid" parameter in addition to video "aid". This cid parameter is returned for API responses, so VocaDB should save it. I'm not sure what is the purpose of that parameter, but it varies by PV.

On bilibili one aid (av12345678) entry can have many episodes, while each of the episode can be referred to with a unique cid. 
See https://api.bilibili.com/x/player/pagelist?aid=12802603 for an example. The video [av12802603](https://www.bilibili.com/video/av12802603) is the full record of the 2017 Luotianyi Concert, and the uploader put each song performed into an individual episode, with respective cids.

AgFlore:
> A problem with the new player is that the playback controls won't appear unless the player is big enough. The embedded player on VocaDB is below that size. I think it's rather silly, because you can't for example seek at all. I wonder if there's an option to force the controls to appear always?

I've tried it out and finds that the threhold of showing controls is that the height of embed must be greater than 360px, sounds like not very far from the currently using 315px. 

Also, according to https://blog.asucreyau.xyz/2018/11/19/bilibili-embed-optimize/, the best height of the embed should be (width*9/16+118px), where the 118px is the total height of the top banner, the time-control and the bottom danmuku input. In the case here the banner and the danmuku input is useless, so perhaps we can mask it out in the final page with some html technique?

PR #572, opened 2020-03-27
Fix Bili playback controls; iframe all Bili player

AgFlore:
The `cid` is in fact unnecessary for Bilibili iframe players (the `page` alone is sufficient for navigating in multi-page videos, see #494 ), and we can still enable the new `<iframe>` players without referring to `cid`s for songs who didn't got their `cid`s here (#495 ).

Bilibili playback controls will show up for `<iframe>`s larger than 480*360 (#372 ). If the frame is larger than the threshold, playback controls (29px in height) will start to show, and another two ad banners (one in the top, 48px in height; the other in the bottom, 38px in height) will also show up. My strategy would be simply masking the two ad banners away, so that we can get the playback controls back while also meeting the vocaDB size requirement (560*315).

### DIFF
--- a/VocaDbWeb/App_Code/PVHelpers.cshtml
+++ b/VocaDbWeb/App_Code/PVHelpers.cshtml
@@ -24,12 +24,16 @@
 	}
 }
 
-@helper EmbedBili(PVContract pv, string widthStr, string heightStr) {
+@helper EmbedBili(PVContract pv, int width, int height) {
 	var meta = pv.ExtendedMetadata != null ? pv.ExtendedMetadata.GetExtendedMetadata<BiliMetadata>() : null;
-	if (meta != null && meta.Cid != 0) {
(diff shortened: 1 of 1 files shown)


riipah:
Weird, I'm pretty sure that I checked cid parameter is required, and the page parameter has always been there. Perhaps they changed it at some point. But yes, I verified that it does indeed work without the cid parameter now.

AgFlore:
Known issues about the commit: 
1. Not showing correctly in the homepage ranking (the size set for the player seemed to be ignored)
2. Not showing correctly in playlists (player became in the same line with the title. Check for example https://vocadb.net/L/8246)

Not very familiar with the HTML stuff, still looking for a better way out.

riipah:
Uh yeah, I was worried about something like that when trying to be creative with the styles. Getting the element to display properly in all situations is difficult. If you can't make it work, then it's better to revert to old layout. Missing the controls isn't that much of a problem.

PR #574, opened 2020-03-28
Update the matcher for Bili user profile & Enable icons for all Bili subdomains

AgFlore:
1. Most newly added Bilibili user profiles for artists go like `http://space.bilibili.com/nnn`. It would be great convenient to have them automatically matched when adding songs.

2. Many major subdomains (even including "http(s)://www.bilibili.com/" itself!) cannot have the icon shown in vocadb's pages, which is ridiculous.

### DIFF
--- a/VocaDbWeb/Content/Styles/ExtLinks.less
+++ b/VocaDbWeb/Content/Styles/ExtLinks.less
@@ -30,7 +30,7 @@ a.extLink {
 		background-image: url("/Content/ExtIcons/atwiki.png");
 	}
 
-	&[href^="http://space.bilibili.tv/"], &[href^="http://www.bilibili.tv/"], &[href^="http://space.bilibili.com/"], &[href^="https://space.bilibili.tv/"], &[href^="https://www.bilibili.tv/"], &[href^="https://space.bilibili.com/"] {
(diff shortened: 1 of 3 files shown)


VocaDbWeb/Content/Styles/ExtLinks.less
```diff
@@ -30,7 +30,7 @@ a.extLink {
 		background-image: url("/Content/ExtIcons/atwiki.png");
 	}
 
-	&[href^="http://space.bilibili.tv/"], &[href^="http://www.bilibili.tv/"], &[href^="http://space.bilibili.com/"], &[href^="https://space.bilibili.tv/"], &[href^="https://www.bilibili.tv/"], &[href^="https://space.bilibili.com/"] {
+	&[href*="bilibili.tv/"], &[href*="bilibili.com/"] {
```
riipah: Maybe href*=".bilibili.tv" (including the dot). Better to make it as specific as possible, if the dot is always there.

AgFlore: Yes, there's no such things as "bilibili.com" so far. The dot is always there, you may add it.

riipah:
Okay, thank you.

Issue #347, opened 2017-12-27
Bilibili cover images do not load


romanrm:
I created a list with a few examples: http://vocadb.net/L/7416
Opening image in a new tab shows that the error is 403 Forbidden.
But copying the URL to a new tab manually (no vocadb HTTP Referrer sent) makes it load fine.
Seems like they blocked loading pictures embedded in external websites?

riipah:
It works in other browsers besides Chrome. In IE at least. I'm not sure if it's a problem with Chrome or Bilibili. I couldn't figure out the reason myself. I know Google isn't in the best of terms with the Chinese government.

riipah:
See related comment here http://vocadb.net/discussion/topics/119

AgFlore:
I guess this may be resolved with replacing the 'https' in the thumb image URL with 'http'.

Embedding with <img>https://i2.hdslb.com/bfs/archive/d6aa7b1bad28372a396d9944bc247ea04a37b976.j
pg</img> may be blocked, while  <img>http://i2.hdslb.com/bfs/archive/d6aa7b1bad28372a396d9944bc247ea04a37b976.j
pg</img> works well.
Tested on other sites, like [this](https://bbs.pku.edu.cn/v2/post-read.php?bid=7&threadid=17518670). Sound speculative, but I think it is worth a try.

riipah:
I'd rather not do that because showing http images on a https site causes warnings for mixed content and may in theory allow cross site attacks. Additionally, that thumbnail URL is coming from Bilibili's own API. I would have to manually "downgrade" it to http, which is silly.

romanrm:
> showing http images on a https site causes warnings for mixed content

And not possible at all soon: https://love2dev.com/blog/chrome-mixed-content/

AgFlore:
Right, I understand. Will try to not place Bilibili PV on the top when editing.

AgFlore:
> 
> 
> I'd rather not do that because showing http images on a https site causes warnings for [mixed content](https://developer.mozilla.org/fi/docs/Web/Security/Mixed_content) and may in theory allow cross site attacks. It will also make the browser classify the website as less secure. I know we still have some http images from nico nico, but hopefully one day we can get rid of them. [There was already someone who was worried about this earlier](https://vocadb.net/discussion/topics/178).
> 
> Additionally, that thumbnail URL is coming from Bilibili's own API. I would have to manually "downgrade" it to http, which is silly.

[...]

PR #579, opened 2020-03-30
Fix displaying problems for Bilibili players from #572 & Fix song thumb loading problems from #347

AgFlore:
For the [problems](https://github.com/VocaDB/vocadb/pull/572#issuecomment-605606181) with the layout of Bilibili players:

> Not showing correctly in playlists (player became in the same line with the title. Check for example https://vocadb.net/L/8246)

Fixed this by adding a style attribute of `width:max-content` to the wrapping `<div>`.

> Not showing correctly in the homepage ranking (the size set for the player seemed to be ignored)
There were multiple problems with the layout in the homepage ranking: 

First, the size set for the `<iframe>` was covered by the stylesheet, so we had to re-address it in the `style` manually. 
Next, setting the `position` as `absolute` (from the stylesheet as well!) would prevent the wrapper from hiding the manually masked content, so I set is as `relative` as in other pages. 
Then we arrrived at the current fix; it's still not perfect, as the player is 25px too low (due to the 25px `padding-top` from `.pvViewerContent .pv-embed-wrapper`'s stylesheet), which ends up to cover the songInfoButton. But I think that's tolerable, as one can check the page for the song by clicking the title, so giving up the button was not that fatal.
Anyway that's not a elegant or perfect fix, still looking forward to getting it better; maybe one need to do some work on the layout/stylesheet of the page as well to fix this completely.

For the issue #347 , one simply add `referrerpolicy="same-origin"` to everywhere a song thumb may be loaded, and hopefully this can be settled. Seems that `referrerpolicy` cannot be used as a `style` attribute, so I cannot fix it in the stylesheets, and I need to add it everywhere manually.

### DIFF
--- a/VocaDbWeb/Views/Album/Details.cshtml
+++ b/VocaDbWeb/Views/Album/Details.cshtml
@@ -356,7 +356,7 @@
 			</h3>
 			<div class="media" data-bind="with: personalDescription">
 				<a data-bind="visible: !author.isEmpty() && author.entry().mainPicture, attr: { href: vdb.utils.EntryUrlMapper.details('Artist', author.id()) }" class="pull-left">
-					<img class="coverPicThumb" alt="Thumb" data-bind="attr: { src: !author.isEmpty() && author.entry().mainPicture ? author.entry().mainPicture.urlSmallThumb : null }" />
+					<img class="coverPicThumb" alt="Thumb" data-bind="attr: { src: !author.isEmpty() && author.entry().mainPicture ? author.entry().mainPicture.urlSmallThumb : null }" referrerpolicy="same-origin" />
 				</a>
 
 				<div class="media-body">
--- a/VocaDbWeb/Views/Album/Details.cshtml
+++ b/VocaDbWeb/Views/Album/Details.cshtml
@@ -356,7 +356,7 @@
 			</h3>
 			<div class="media" data-bind="with: personalDescription">
 				<a data-bind="visible: !author.isEmpty() && author.entry().mainPicture, attr: { href: vdb.utils.EntryUrlMapper.details('Artist', author.id()) }" class="pull-left">
-					<img class="coverPicThumb" alt="Thumb" data-bind="attr: { src: !author.isEmpty() && author.entry().mainPicture ? author.entry().mainPicture.urlSmallThumb : null }" referrerpolicy="same-origin" />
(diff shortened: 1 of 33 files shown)


VocaDbWeb/Views/Album/Details.cshtml
```diff
@@ -356,7 +356,7 @@
 			</h3>
 			<div class="media" data-bind="with: personalDescription">
 				<a data-bind="visible: !author.isEmpty() && author.entry().mainPicture, attr: { href: vdb.utils.EntryUrlMapper.details('Artist', author.id()) }" class="pull-left">
-					<img class="coverPicThumb" alt="Thumb" data-bind="attr: { src: !author.isEmpty() && author.entry().mainPicture ? author.entry().mainPicture.urlSmallThumb : null }" />
+					<img class="coverPicThumb" alt="Thumb" data-bind="attr: { src: !author.isEmpty() && author.entry().mainPicture ? author.entry().mainPicture.urlSmallThumb : null }" referrerpolicy="same-origin" />
```
riipah, line 359: Why would you add it here? Album images are always from VocaDB domain

VocaDbWeb/Views/Search/Partials/_TagSearchList.cshtml
```diff
@@ -34,7 +34,7 @@
 			<tr>
 				<td style="width: 80px">
 					<a data-bind="visible: $data.mainPicture && mainPicture.urlSmallThumb, attr: { href: vdb.utils.EntryUrlMapper.details_tag(id, urlSlug), title: $data.additionalNames }" href="#">
-						<img data-bind="attr: { src: ($data.mainPicture ? mainPicture.urlSmallThumb : '') }" title="Cover picture" class="coverPicThumb img-rounded" />
+						<img data-bind="attr: { src: ($data.mainPicture ? mainPicture.urlSmallThumb : '') }" title="Cover picture" class="coverPicThumb img-rounded" referrerpolicy="same-origin" />
```
riipah, line 37: But surely these images are never off-site?

AgFlore: Yeps, I was doing quite blindly when adding the referrerpolicy tag as I was not 100% definite about what each picture actually is. Supposingly only song thumbs need this; but with the `same-origin` attribution, images already from our own domain wouldn't be effected, so over-tagging would only be a tidiness issue. I've removed the two unnecessary occurances you mentioned.

riipah:
All right, thank you.

Issue #564, opened 2020-03-04
Prevent multiple account creation with email aliases


riipah:
Recently I've seen spammers creating multiple accounts using email aliases, that is user+1@gmail.com, user+2@gmail.com, user+3@gmail.com etc. IP is different, so blocking IP doesn't help.

We should either add option to ban email address, with aliases, or prevent creating multiple accounts (including disabled) with the same email, ignoring alias.

That is, either we ban user+*@gmail.com or we consider user+1@gmail.com and user+2@gmail.com the same so that only one of them can exist. I think the latter one makes more sense and should be easier to implement too.

ycanardeau:
We have to consider not only user+1@gmail.com, user+2@gmail.com and etc., but also us.er@gmail.com, u.s.e.r@gmail.com and etc. the same because [dots don't matter in Gmail addresses](https://support.google.com/mail/answer/7436150).

riipah:
Good point, I didn't know that. I've only seen the + aliases. I think the dots will be more difficult to check...

ycanardeau:
Can be closed now?

PR #580, opened 2020-03-30
Feat/564 normalize email addresses

ycanardeau:
#564 I ported [iDoRecall:email-normalize](https://github.com/iDoRecall/email-normalize) to C# by using [DnsClient.NET](https://github.com/MichaCo/DnsClient.NET). However, if the [DetectProvider](https://github.com/ycanardeau/vocadb/blob/3aad17a56ffa8209c5a9a7502468adb49a38b995/VocaDbModel/Helpers/MailAddressNormalizer.cs#L16) flag is set as the second argument of Normalize or NormalizeAsync, the program freezes at [LookupClient.QueryAsync](https://github.com/ycanardeau/vocadb/blob/3aad17a56ffa8209c5a9a7502468adb49a38b995/VocaDbModel/Helpers/MailAddressNormalizer.cs#L121) for some reason, even though their tests have passed. Additionally, we have to update the NormalizedEmail field for existing users somehow.

### DIFF
--- a/Tests/Helpers/MailAddressNormalizerTests.cs
+++ b/Tests/Helpers/MailAddressNormalizerTests.cs
@@ -1,0 +1,86 @@
+using System.Threading.Tasks;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using VocaDb.Model.Helpers;
+
+namespace VocaDb.Tests.Helpers {
+
+	/// <summary>
+	/// Code from https://github.com/iDoRecall/email-normalize/blob/0938e0a4710c6fc076c50dd42ea2886b2984e219/tests.js
+	/// </summary>
+	[TestClass]
+	public class MailAddressNormalizerTests {
+
+		private async Task<string> NormalizedAddress(string address, MailAddressNormalizerOptions options = MailAddressNormalizerOptions.None) {
+			return await MailAddressNormalizer.NormalizeAsync(address, options);
+		}
+
+		[TestMethod]
+		public async Task OnlySupportedDomains() {
+			Assert.AreEqual("a.b.c+tag@example.com", await NormalizedAddress("a.b.c+tag@example.com"));
+		}
+
+		[TestMethod]
+		public async Task GmailDots() {Assert.AreEqual("abc@gmail.com", await NormalizedAddress("a.b.c@gmail.com"));
+			Assert.AreEqual("a.b.c@yahoo.com", await NormalizedAddress("a.b.c@yahoo.com"));
+		}
+
+		[TestMethod]
+		public async Task Plus() {
+			Assert.AreEqual("abc@gmail.com", await NormalizedAddress("a.b.c+tag@gmail.com"));
+			Assert.AreEqual("a.b.c+tag@yahoo.com", await NormalizedAddress("a.b.c+tag@yahoo.com"));
+		}
+
+		[TestMethod]
+		public async Task NonStandardTlds() {
+			Assert.AreEqual("a.b.c+tag@something.co.uk", await NormalizedAddress("a.b.c+tag@something.co.uk"));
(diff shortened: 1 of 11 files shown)


Tests/Helpers/MailAddressNormalizerTests.cs
```diff
@@ -0,0 +1,86 @@
...
+	[TestClass]
+	public class MailAddressNormalizerTests {
+
+		private async Task<string> NormalizedAddress(string address, MailAddressNormalizerOptions options = MailAddressNormalizerOptions.None) {
+			return await MailAddressNormalizer.NormalizeAsync(address, options);
+		}
+
+		[TestMethod]
+		public async Task OnlySupportedDomains() {
+			Assert.AreEqual("a.b.c+tag@example.com", await NormalizedAddress("a.b.c+tag@example.com"));
```
riipah, line 19: You could simplify the tests further by extracting TestNormalizedEmail method, taking expected and given email
```
public void TestNormalizedEmail(string expected, string given) { 
    Assert...
}
```

VocaDbModel/Helpers/MailAddressNormalizer.cs
```diff
@@ -0,0 +1,136 @@
...
+using DnsClient;
+
+namespace VocaDb.Model.Helpers {
+
+	[Flags]
+	public enum MailAddressNormalizerOptions {
+		None = 0,
+		ForceRemoveDots = 1,
+		ForceRemoveTags = 2,
+		DetectProvider = 4
```
riipah, line 16: Please add comment here

VocaDbModel/Helpers/MailAddressNormalizer.cs
```diff
@@ -0,0 +1,136 @@
...
+					user = user.Split(separator)[0];
+			}
+
+			if (options.HasFlag(MailAddressNormalizerOptions.ForceRemoveDots) || Regex.IsMatch(host, @"^(gmail|googlemail|google)\.com$"))
+				user = Regex.Replace(user, @"\.", "");
+
+			if (host == "googlemail.com")
+				host = "gmail.com";
+
+			if (options.HasFlag(MailAddressNormalizerOptions.DetectProvider)) {
```
riipah, line 107: I think we don't actually need this, for now at least. It's unlikely the spammers would go through the trouble of custom domain, at least so far they haven't. It's fine to keep it simple for now. In my opinion, it might be better remove the code if it doesn't even work properly.

VocaDbModel/Domain/Users/User.cs
```diff
@@ -358,6 +361,17 @@ public User(string name, string pass, string email, IPasswordHashAlgorithm passw
 			}
 		}
 
+		/// <summary>
+		/// Normalized email address.
```
riipah, line 365: Maybe add an explanation what it means (Email address without "+" or dots).

riipah:
Thank you again. I think the DetectProvider code isn't needed right now. We can add it later if needed.

As for adding NormalizedEmail to users, we have 3 options:
* Simply copy Email address field for current users (it could be enough), even though it's not normalized
* Do the transformation in the migration (you can refer the same code)
* Add a piece of code to AdminController. I've added similar migrations there before.

VocaDbModel/Service/AdminService.cs
```diff
@@ -594,6 +594,25 @@ public class AdminService : ServiceBase {
 
 		}
 
+		public void UpdateNormalizedEmailAddresses() {
+
+			PermissionContext.VerifyPermission(PermissionToken.Admin);
```
ycanardeau, line 599: What should I pass for the argument here?

riipah: This is ok.

riipah:
Thanks, it looks fine now.

Issue #571, opened 2020-03-25
Add URL matcher for the new Bilibili `BVID`


AgFlore:
Bilibili has started [a new video-identifier system](https://www.bilibili.com/read/cv5167957/) called `bvid` featuring a string starting with "BV". The old "av1234567" `aid` identifer is still accessible, but no longer the default video URL pattern. 
For example, https://www.bilibili.com/video/av6009789 is now equivalent (and usually redirected to) https://www.bilibili.com/video/BV1Qs411k7Qv

Both `aid` and `bvid` identifiers are accessible from the api currently used, for example https://api.bilibili.com/x/web-interface/archive/stat?bvid=1Qs411k7Qv and https://api.bilibili.com/x/web-interface/archive/stat?aid=6009789 returns the same. 

Generally, the `bvid` can be [generated](https://gist.github.com/abc1763613206/b8afbb88849835f064f74e652fdb16c5) from `aid`; basically this involves writing the `aid` under base58 and `xor`-ing it with a magic number.

Since there's no tracks of discarding the `aid` identifier, currently I see no need in migrating our database from `aid` to `bvid`; adding a URL matcher shall be sufficient.

(And also about the video-thumb issue: I think adding a `referrerpolicy="no-referrer"` to our `img` will fix.)

riipah:
To save my time, could you check the [API documentation](https://github.com/VocaDB/vocadb/wiki/Bilibili-API-documentation) and see how it has changed? Does the "Video info request" now return bvid?

AgFlore:
http://api.bilibili.com/view?type=json&appkey=8e9fc618fbd41e28&id=6009789 (6009789 is `aid`) works as usual, and replacing it with bvid doesn't work.

https://api.bilibili.com/x/web-interface/archive/stat?bvid=1Qs411k7Qv now returns:
`{"code":0,"message":"0","ttl":1,"data":{"aid":6009789,"bvid":"BV1Qs411k7Qv","view":1966272,"danmaku":57008,"reply":107610,"favorite":127873,"coin":97931,"share":16065,"like":43061,"now_rank":0,"his_rank":28,"no_reprint":0,"copyright":1,"argue_msg":"","evaluation":""}}`

riipah:
I see. Do you know if they are planning on replacing that "Video info request"? And should we still use the "Player request" to get the duration (also called using aid)?

One way to do this is to save both aid and bvid, similar as what we've done with SoundCloud.

[...]

PR #581, opened 2020-04-02
Feat/571 bilibili antibv

ycanardeau:
Workaround for #571. I ported [this code](https://www.zhihu.com/question/381784377/answer/1099438784) to C#. I'm not sure whether or not this works correctly because I don't have an API key. This is very hacky and should be refactored.


VocaDbModel/Helpers/BilibiliHelper.cs
```diff
@@ -0,0 +1,44 @@
+using System;
+using System.Collections.Generic;
+
+namespace VocaDb.Model.Helpers {
+
+	/// <summary>
+	/// Code from: https://www.zhihu.com/question/381784377/answer/1099438784
+	/// </summary>
+	public static class BilibiliHelper {
```
riipah, line 9: Maybe add a simple description what this does.

VocaDbModel/Service/VideoServices/VideoServiceBilibili.cs
```diff
@@ -53,18 +54,19 @@ public VideoServiceBilibili()
 		public override async Task<VideoUrlParseResult> ParseByUrlAsync(string url, bool getTitle) {
 
 			var id = GetIdByUrl(url);
+			var aid = id.StartsWith("BV") ? BilibiliHelper.Decode(id).ToString() : id;
```
riipah, line 57: We should definitely save both aid and bvid. Bvid can be added to BiliMetadata, same as Cid.

VocaDbModel/Service/VideoServices/VideoServiceBilibili.cs
```diff
@@ -53,18 +54,19 @@ public VideoServiceBilibili()
...
-				return VideoUrlParseResult.CreateOk(url, PVService.Bilibili, id, VideoTitleParseResult.Empty);
+				return VideoUrlParseResult.CreateOk(url, PVService.Bilibili, aid, VideoTitleParseResult.Empty);
 			}
 
-			var paramStr = string.Format("appkey={0}&id={1}&type=json{2}", AppConfig.BilibiliAppKey, id, AppConfig.BilibiliSecretKey);
+			var paramStr = string.Format("appkey={0}&id={1}&type=json{2}", AppConfig.BilibiliAppKey, aid, AppConfig.BilibiliSecretKey);
 			var paramStrMd5 = CryptoHelper.HashString(paramStr, CryptoHelper.MD5).ToLowerInvariant();
 
-			var requestUrl = string.Format("https://api.bilibili.com/view?appkey={0}&id={1}&type=json&sign={2}", AppConfig.BilibiliAppKey, id, paramStrMd5);
+			var requestUrl = string.Format("https://api.bilibili.com/view?appkey={0}&id={1}&type=json&sign={2}", AppConfig.BilibiliAppKey, aid, paramStrMd5);
```
riipah, line 69: Actually, we should replace this to use the new API, as discussed in #581. Apparently the new API doesn't even need an API key, so you can test it properly, and no need to do the hacky encoding/decoding either.

VocaDbModel/Service/VideoServices/VideoServiceBilibili.cs
```diff
@@ -61,37 +42,37 @@ public VideoServiceBilibili()
...
 
-			var title = HtmlEntity.DeEntitize(response.Title);
-			var thumb = response.Pic ?? string.Empty;
-			var author = response.Author ?? string.Empty;
-			var created = response.CreatedAt;
-			var length = await GetLength(id);
+			var title = HtmlEntity.DeEntitize(response.Data.Title);
+			var thumb = response.Data.Pic ?? string.Empty;
+			var author = response.Data.Owner.Name ?? string.Empty;
+			var created = response.Data.PubDate;
```
ycanardeau, line 67: The publication date and time are represented in UTC, but it is displayed as local time (UTC+10:00 in my case, my clock is one hour ahead of JST, Japan Standard Time, by the way) in the song edit page. So, assuming that pubdate is 1579877678 (1/24/2020 14:54:38 in UTC), it is displayed as 1/25/2020 in my timezone, even though it is correctly stored as 1/24/2020 in the database. The same problem occurs for other PV services like SoundCloud.

VocaDbModel/Service/VideoServices/VideoServiceBilibili.cs
```diff
@@ -61,37 +42,37 @@ public VideoServiceBilibili()
 				return VideoUrlParseResult.CreateOk(url, PVService.Bilibili, id, VideoTitleParseResult.Empty);
```
ycanardeau, line 42: I think we should use aid instead of id here. Can we do this after getting a BilibiliResponse?

VocaDbModel/Service/VideoServices/VideoServiceBilibili.cs
AgFlore: For Bilibili the case is easier, most of its producers and users live in GMT+8, so I think it would be ok to fix it to the one timezone when filling the timezone. (Anyway generally this one-day difference doesn't matter on vocadb.)

VocaDbModel/Service/VideoServices/VideoServiceBilibili.cs
```diff
@@ -110,26 +91,31 @@ public VideoServiceBilibili()
 
 	[DataContract(Namespace = Schemas.VocaDb)]
 	public class BiliMetadata {
+		[DataMember]
+		public string Bvid { get; set; }
 		[DataMember]
 		public int Cid { get; set; }
 	}
```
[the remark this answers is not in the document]
AgFlore: I would suggest do that. It is not impossible that one day they give up `aid` and move entirely to `bvid`, so storing `aid` as well in the metadata may help ease the problem then.

VocaDbModel/Service/VideoServices/VideoServiceBilibili.cs
```diff
@@ -110,26 +91,31 @@ public VideoServiceBilibili()
...
 
-	class PlayerResponse {
-		public PlayerResponseData[] Data { get; set; }
+	class BilibiliResponse {
+		public BilibiliResponseData Data { get; set; }
 	}
 
-	class PlayerResponseData {
+	class BilibiliResponseData {
+		public int Aid { get; set; }
```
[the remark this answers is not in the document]
AgFlore: The `bvid` was implemented just as `aid` was about to reach 1e8. After the implementation of `bvid`, the `aid` became no longer ordinal; new videos would be assigned a random number as its `aid`, currently ranging up to 1e9 (still within `int`). I would suggest keep what we had used for `aid` to maintain consistency.

VocaDbModel/Helpers/HtmlRequestHelper.cs
```diff
@@ -34,10 +34,12 @@ public static class HtmlRequestHelper {
 				response.Close();
 			}
 
-		}
+		}
+
+		public static Task<T> GetStreamAsync<T>(string url, Func<Stream, T> func) => GetStreamAsync(url, func, TimeSpan.FromSeconds(10000));
```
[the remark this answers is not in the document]
riipah: Hmm, don't know why it's so long. Maybe 300 sec would be enough.

VocaDbModel/Service/VideoServices/VideoServiceBilibili.cs
riipah: Does the response contain timezone information? The problem is that DateTime uses local timezone, whatever that is. Timezone can be saved by using DateTimeOffset. I recommend saving the date in database as UTC. But as AgFlore said, it doesn't really matter that much.

VocaDbModel/Service/VideoServices/VideoServiceBilibili.cs
riipah: Not sure what you mean.

VocaDbModel/Service/VideoServices/VideoServiceBilibili.cs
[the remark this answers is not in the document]
riipah: Doesn't make much difference either way. That metadata aid field won't be available for older PVs. aid is already saved in the main ID field. On the other hand, it won't use much memory saving it here.

riipah:
Are the "new" aids going to be unique still, or could a new PV get the same aid as an existing PV, if they are randomized?

VocaDbModel/Service/VideoServices/VideoServiceBilibili.cs
riipah: Ok yeah, I see it now. Because here you're returning id, but below you're returning aid. Actually, I'm thinking, could you just always return the given id, either aid or bvid?

AgFlore:
> 
> 
> Are the "new" aids guaranteed to be unique still, or could a new PV get the same aid as an existing PV?

At least for now they are still unique, `aid` is still a video identifier, and one can access any video with the `avnnnnnnnn` URL. Breaking the uniqueness would mean totally discarding `aid`s  even for Bilibili themselves, and I don't think they would do that for consistency reasons.

VocaDbModel/Service/VideoServices/VideoServiceBilibili.cs
ycanardeau: > Are you talking about always returning aid instead of id (which may be bvid)?

Yes.

> And would you use the Encode/Decode code to get the ID?

No, we don't need to use the Encode/Decode code to get the ID ([I've already removed it](https://github.com/VocaDB/vocadb/pull/581/commits/46b725052ae075eec6c01f8c1c3d13e7098f1a5f)). Instead, we can use the `aid` and `bvid` fields included in a `BilibiliResponse`. That's why I want to move this if block after the try-catch block.

> Am I correct that this ID always includes the prefix (av or bv), so we can always figure out whether it's aid or bvid?

The `id` variable may be either aid or bvid as you said. If the `id` variable is an aid, then it is just an integer, that is, without the `av` prefix. For example, `6009789`. If the `id` variable is a bvid, then it starts with the `BV` prefix. For example, `BV1Qs411k7Qv`.

riipah: Ideally, we would avoid any web requests if getTitle is false. It's a performance optimization, because sometimes we only need the ID and not metadata, and the request might be made in a loop for multiple PVs. But since Bilibili isn't used as much as YouTube or Nico Nico, I guess it's not a big issue. Some other parsers such as Piapro also make a web request for parsing ID, because they have to. If you're going that route, then you might as well just remove this if check, right?

riipah:
Thanks, it should be good now

Issue #16, opened 2015-01-18
Moving all entry images under the static domain


riipah:
Album/artist additional images and tag/songlist main images are saved under the application folder, they should be moved under the static domain, like album/artist main thumbnails are already.

Code changes are mostly done, the actual migration work remains.

- [x] album/artist additional images
- [x] tag/songlist images

PR #582, opened 2020-04-04
Update views

ycanardeau:


VocaDbWeb/Views/SongList/Details.cshtml
```diff
@@ -88,12 +88,20 @@
 						<a href="@Url.EntryDetails(ev, ev.UrlSlug)">
 							@ev.Name
 						</a>
-						@if (ev.Date.HasValue && !string.IsNullOrEmpty(ev.VenueName)) {
-							<small>(@ev.Date.Value.ToString("d"), @ev.VenueName)</small>
+						@if (ev.Date.HasValue) {
+							if (ev.Venue != null) {
+								<small>(@ev.Date.Value.ToString("d"), <a href="@Url.EntryDetails(ev.Venue)">@ev.Venue.Name</a>)</small>
```
riipah, line 93: I'm thinking you could make choosing venue entry vs. venue name into some sort of helper (maybe in EventHelpers.cshtml), because it's used in many places.

ycanardeau: Yeah, I was thinking just the same thing. I thought it would be better to do that after replacing SongListDetailsContract with SongListForApiContract.

VocaDbModel/DataContracts/ReleaseEvents/ReleaseEventForApiContract.cs
```diff
@@ -103,6 +103,8 @@ public class ReleaseEventForApiContract : IReleaseEvent, IEntryBase {
 		[DataMember]
 		public DateTime? EndDate { get; set; }
 
+		public bool HasVenueOrVenueName => Venue != null || !string.IsNullOrEmpty(VenueName);
```
ycanardeau, line 106: Is it possible for an API contract to have read-only properties?

riipah: There should be no problem with that.

VocaDbWeb/Views/SongList/Details.cshtml
ycanardeau: Perhaps it would be better to replace SongListDetailsContract with SongListForApiContract after moving all entry images under the static domain (#16).

riipah: You could also just add interfaces, similar to IUser. It might be a good idea anyway.

ycanardeau: Can I add a new property for URL to original picture to EntryThumbForApiContract? If not, how can I show the original picture on the song list details page? Perhaps we should not replace SongListDetailsContract with SongListForApiContract for now. (Additionally, we'll need some new interfaces and classes like ISongList and IUserExtender. By the way, how about renaming `-Extender` classes to `-Extensions` after merging #583?)

riipah: I'll try to [refactor image URL handling a bit](https://github.com/VocaDB/vocadb/commit/6e043d735b540ab7e68ca258de413b17d7604167), so we could add that original image URL to EntryThumbForApiContract.

riipah: Huh, I refactored the image URL generation, which ended up being a bit more than expected (which is common). Anyway, EntryThumbForApiContract now has original image URL, and the URL generation is much more straightforward in general. Just use the "aggregate" image URL factory.

ycanardeau: Thank you! I'll try to update this PR in the next few days.

VocaDbModel/Domain/Users/IUser.cs
```diff
@@ -10,5 +10,21 @@ public interface IUser {
...
+		public static bool IsTheSameUser(this IUser left, IUser right) {
+
+			ParamIs.NotNull(() => left);
+
+			if (right == null)
+				return false;
+
+			return left.Id == right.Id;
+
+		}
```
ycanardeau, line 26: Can I convert IsTheSameUser to an extension method like this?

VocaDbModel/Domain/Songs/ISongList.cs
```diff
@@ -0,0 +1,13 @@
...
+
+namespace VocaDb.Model.Domain.Songs {
+
+	public interface ISongList {
+
+		IUser Author { get; }
+
+		bool FeaturedList { get; }
+
+	}
```
ycanardeau, line 11: I added ISongList so that EntryPermissionManager.CanEdit can be called from SongListDetailsViewModel.

VocaDbModel/Database/Queries/SongListQueries.cs
```diff
@@ -301,7 +303,7 @@ public class SongListQueries : QueriesBase<ISongListRepository, SongList> {
 					list = ctx.Load(contract.Id);
 					var diff = new SongListDiff();
 
-					EntryPermissionManager.VerifyEdit(PermissionContext, list);
+					EntryPermissionManager.VerifyEdit(PermissionContext, (ISongList)list);
```
riipah, line 306: Why is this cast needed?

VocaDbModel/Domain/Users/IUser.cs
riipah: I think if you add IEquatable<IUser> to IUser then this should not be necessary.

VocaDbModel/Database/Queries/SongListQueries.cs
ycanardeau: Because the call is ambiguous between `EntryPermissionManager.CanEdit(IUserPermissionContext, ISongList)` and `EntryPermissionManager.CanEdit(IUserPermissionContext, IEntryWithStatus)`. Actually, SongList is enough for EntryPermissionManager.VerifyEdit.

riipah: Then I recommend that you should add CanEdit overloads for the concrete types so that there is no ambiguity. Or rename it to CanEditSongList. I would prefer getting rid of the casts if possible.

riipah:
Looks good now, thanks

PR #583, opened 2020-04-05
Migrate to package references and SDK project format

ycanardeau:
#550

References:
https://www.michaeltaylorp3.net/migrating-to-sdk-project-format/
https://www.michaeltaylorp3.net/migrating-to-package-references/
dotnet/standard#542


VocaDbModel/Mapping/Security/PermissionTokenMap.cs
```diff
@@ -1,18 +0,0 @@
...
-
-	public class PermissionTokenMap : ClassMap<PermissionToken> {
-
-		public PermissionTokenMap() {
-
-			Id(m => m.Id);
-
-		}
-	}
-}
```
ycanardeau, line 18: I removed PermissionTokenMap.cs because it seems that it is not used. In fact, if it is included in the project, an NHibernate.InvalidProxyTypeException will be thrown.

VocaDbWeb/Web.config
```diff
@@ -41,6 +41,8 @@
   <system.web>
     <compilation debug="true" targetFramework="4.8" batch="false">
       <assemblies>
+        <!-- https://github.com/dotnet/standard/issues/542 -->
+        <add assembly="netstandard, Version=2.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51" />
```
riipah, line 45: Did you get the error? I was under the impression this should be unnecessary if AutoGenerateBindingRedirects and PackageReference are enabled.

ycanardeau: You're right. Reverted Web.config.

ycanardeau:
Thank you for your review!

riipah:
Thank you for the PR!

PR #585, opened 2020-04-10
Update Linkmatchers for 5sing, NCM, lofter and moegirlpedia

AgFlore:
5sing had migrated its domain from www.5sing.com to 5sing.kugou.com long long ago.
Also added link matchers for NetEase Cloud Music (a Chinese music streaming service, widely used for song releases), Lofter (a Chinese blog service) and Moegirlpedia. Updated their icons as well.


AgFlore:
And also added two unaddressed thumb referrer policy unaddressed in #579  (one in the activity feed, the other in playlists).

VocaDbWeb/Content/Styles/ExtLinks.css
```diff
@@ -295,3 +295,15 @@ a.extLink[href^="http://vocaloid.wikia.com/wiki/"],
 a.extLink[href^="https://vocaloid.fandom.com/wiki/"] {
   background: url("/Content/ExtIcons/vocawiki.png") no-repeat left 50%;
 }
+a.extLink[href*=".moegirl.org/"] {
```
riipah, line 298: Please understand that these wildcard matchers have worse performance than prefix matchers. Are all of these required to be wildcards?

AgFlore: > Please understand that these wildcard matchers have worse performance than prefix matchers. Are all of these required to be wildcards? If it's just http/https, then better have it has prefix matcher.

I see. Modified some of them.

riipah:
All right, it looks fine now, thanks!

VocaDbWeb/Content/Styles/ExtLinks.css
riipah: Well that, and it's also safer to make it as specific as possible.

PR #592, opened 2020-04-25
Sort by venue

ycanardeau:


VocaDbModel/Service/QueryableExtenders/ReleaseEventQueryableExtender.cs
```diff
@@ -150,7 +165,9 @@ public enum EventSortRule {
 
 		AdditionDate,
 
-		SeriesName
+		SeriesName,
+
+		Venue
```
riipah, line 170: Please change this to "VenueName"

PR #604, opened 2020-07-12
Fix Overrides.less

ycanardeau:
I removed redundant CSS files because they can be replaced with Less files.


riipah:
Thanks, looks fine!

PR #617, opened 2020-09-12
Update Bilibili player outlay: remove top banner mask

AgFlore:
The Bilibili iframe player has removed the top banner ads, so we remove the obsolete -48px top margin.


riipah:
Hmm okay, I'll try it.

PR #619, opened 2020-09-18
Artist page should not show events too much into future

ycanardeau:


VocaDbModel/Service/Queries/ArtistRelationsQuery.cs
```diff
@@ -57,6 +57,7 @@ public class ArtistRelationsQuery {
 			return session.Query<ReleaseEvent>()
 				.WhereNotDeleted()
 				.Where(e => e.AllArtists.Any(a => a.Artist.Id == id))
+				.WhereDateIsBetween(begin: null, end: DateTime.Now.AddMonths(6))
 				.OrderByDate(SortDirection.Descending)
```
ycanardeau, line 61: How about using `SortDirection.Ascending` instead of `SortDirection.Descending`?

riipah: Hmm, I think here we would want to show past events too, if there is no future events. It might need two queries, like on the front page "new and upcoming albums", but it's ok if it's cached. But this end: Today + 6 months filter is better than old one.

PR #624, opened 2020-09-27
Replace MvcHtmlString with IHtmlString

ycanardeau:


VocaDbWeb.Core/VocaDbWeb.Core.csproj
```diff
@@ -5,18 +5,40 @@
...
 
   <ItemGroup>
+    <Folder Include="Code\Markdown\" />
     <Folder Include="wwwroot\js\" />
     <Folder Include="wwwroot\lib\" />
   </ItemGroup>
 
   <ItemGroup>
     <PackageReference Include="Autofac.Extensions.DependencyInjection" Version="6.0.0" />
+    <PackageReference Include="MarkdownSharp" Version="2.0.5" />
```
ycanardeau, line 41: Installed `MarkdownSharp`.

VocaDbWeb/VocaDbWeb.csproj
```diff
@@ -42,7 +42,7 @@
     <DebugType>full</DebugType>
     <Optimize>false</Optimize>
     <OutputPath>bin\</OutputPath>
-    <DefineConstants>DEBUG;TRACE</DefineConstants>
+    <DefineConstants>TRACE;DEBUG;NETFRAMEWORK</DefineConstants>
```
ycanardeau, line 45: The `NETFRAMEWORK` constant must be removed after we finished migrating to .NET Core.

VocaDbWeb/VocaDbWeb.csproj
```diff
@@ -54,7 +54,7 @@
     <DebugType>pdbonly</DebugType>
     <Optimize>true</Optimize>
     <OutputPath>bin\</OutputPath>
-    <DefineConstants>TRACE</DefineConstants>
+    <DefineConstants>TRACE;NETFRAMEWORK</DefineConstants>
```
ycanardeau, line 57: here too

riipah:
Is this ready for review or are you still working on it?

ycanardeau:
I came up with three different ways.
1. Simply copy files from `VocaDbWeb` to `VocaDbWeb.Core`. It's difficult to manage both of them at the same time.
1. Make interfaces and/or abstract classes for `UrlHelper` (ASP.NET) and `IUrlHelper` (ASP.NET Core), for example.
1. Use preprocessor directives depending on the target platform.
I took the last one, and all classes in the PR has worked fine in an experiment. I'm not sure if this way works fine for other classes.

riipah:
Oh okay hmm. Actually I was thinking about doing 1) because it's the simplest, but yes, then you need to make sure not to make changes to those pages that are already moved. For some pages sharing the page with preprocessor directives or interfaces makes sense, if it's not too much. So I guess choose whichever works best depending on the page.

VocaDbWeb/Code/DynamicImageUrlFactory.cs
```diff
@@ -1,5 +1,10 @@
 using System;
+#if NETFRAMEWORK
```
riipah, line 2: I would actually recommend making a new project (VocaDbWeb.Shared) for these. It's complicated if VocaDbWeb.Core is referring to VocaDbWeb.

riipah:
Okay, I only now understood what this is about. So you're asking what to do about the .cs (code) files in the web project, that have dependencies to ASP.NET.

Would it be possible to move these files to a separate project/folder? Even if it's not a Visual Studio project, having the shared files outside of VocaDb.Web would make more sense. If it's not too complicated, I would recommend actually doing option 2) and writing those wrapper interfaces, as was done for VocaDb.Model.

There is also something called "Shared Projects" (shproj) which is made for use cases like this (although I believe it's mostly used for Xamarin/Universal apps, mobile development and such). For example, [ASP.NET OData uses shproj](https://github.com/OData/WebApi/blob/master/src/Microsoft.AspNet.OData.Shared/Microsoft.AspNet.OData.Shared.shproj) to share code between ASP.NET MVC and Core. But I don't know how complicated that is to set up.

ycanardeau:
And we can replace some static classes with non-static ones, because [ASP.NET Core supports dependency injection into views](https://docs.microsoft.com/en-us/aspnet/core/mvc/views/dependency-injection?view=aspnetcore-3.1).

PR #631, opened 2020-10-16
Embed Twitter timeline

ycanardeau:


VocaDbWeb/Views/Shared/Partials/_LeftMenu.cshtml
```diff
@@ -76,5 +76,7 @@
 			<hr />
 			Html.RenderPartial("Partials/_PaypalDonateButton");
 		}
-	</div>						
+	</div>
+
+	<a class="twitter-timeline" data-height="520" href="https://twitter.com/VocaDB">Tweets by VocaDB</a> <script async src="https://platform.twitter.com/widgets.js" charset="utf-8"></script>
```
riipah, line 81: Should the script be at the end of body as usual, or was it instructed to be put right here? Do you think the left menu is wide enough for the twitter feed? I was thinking it could also be on the front page, maybe right panel after comments.

ycanardeau: I've generated it with [Twitter Publish](https://publish.twitter.com/), so I think it's okay for now. I've also tested it on the top page, but it looks not so good. Should we hide it on the registration and login pages in some way?

riipah:
Let's try it like this for now

Issue #598, opened 2020-06-15
Add ASP.NET Core web frontend


riipah:
Part of #544.

Convert VocaDB.Web project to ASP.NET Core / .NET Core.

Probably the easiest way is to add a new ASP.NET Core project, then copy all content and convert it page per page.

The biggest work is probably replacing the Razor helpers. The syntax is different in ASP.NET Core: https://www.mikesdotnetting.com/article/344/what-happened-to-helpers-in-asp-net-core, 
but otherwise it should be straightforward.

* Global.asax (move to Startup.cs)
* DI container (AutoFac) can be used still, or replace with the Microsoft container.
* Application logic from controllers should be moved to Model.
* Converting Razor helpers
* Configuration access

riipah:
Because it's impossible to do at once, I created a new project that will be converted in parts. In the future branch, the old and new web fronts can be run side by side.

PR #632, opened 2020-10-17
Convert helper to partial view

ycanardeau:
#598 With the help of @bbepis and @riipah. There are a lot of commits in this PR. Should we use `Squash and merge` rather than `Create a merge commit`?


riipah:
I didn't look at the files so closely, I assume this it's mostly copied from helpers to partials. Looks fine though. And this merge commit is fine.

ycanardeau:
Thanks!

PR #633, opened 2020-10-24
Use the robots meta tag

ycanardeau:
Inspired by [MediaWiki](https://www.mediawiki.org/wiki/MediaWiki).


VocaDbWeb/Views/Album/Details.cshtml
```diff
@@ -16,6 +16,8 @@
 @{
 	ViewBag.Parents = new[] { Html.ActionLink(ViewRes.SharedStrings.Albums, "Index") };
 	var url = PageProperties.CanonicalUrl;
+
+	PageProperties.Robots = Model.Deleted ? "noindex,follow" : string.Empty;
```
riipah, line 20: Can you add a named constant for these?

riipah:
Please add the named constant for the "robots" values at least (I think it's the same everywhere). And consider moving them to controller instead of view, if possible.

VocaDbWeb/Views/Album/Details.cshtml
ycanardeau: What would you name those constants? We have two different values, `noindex,follow` and `noindex,nofollow`. And where would you place them?

ycanardeau:
Thanks for your review! I considered moving them to controller. but some methods like `UserController.Login` and `UserController.Create` have multiple overloads, so I think it's better not to do that.

VocaDbWeb/Views/Album/Details.cshtml
riipah: Maybe you could place them in the PagePropertiesData class? The name could be just ```Robots_NoIndex_Follow``` (I think it's still better to have named constant, even if the constant includes the whole string)

ycanardeau:
Added named constants.

riipah:
Thanks, it's ok now.

Issue #639, opened 2020-10-31
Missing implementation for ignoreIds parameter in /api/albums/top


TheCardinalSystem:
**Website**
VocaDB

**Describe the bug**
The API is missing an implementation for the `ignoreIds` parameter in `/api/albums/top`. In the documentation it requests that `ignoreIds` is passed as the body of a POST request, however the endpoint only accepts GET requests

I attempted passing the values as URL parameters and headers, but to no avail. Interestingly enough, the documentation says that `ignoreIds` is a *required* field.

**To Reproduce**
1. Go to [https://vocadb.net/swagger/ui/index](https://vocadb.net/swagger/ui/index)
2. Click on AlbumApi
3. Click on `/api/albums/top`
4. Add values to `ignoreIds` text field
5. Click "Try it out!"
6. Repeat steps 4 and 5 with different values

**Expected behavior**
The result is expected to exempt albums with ids matching those in the `ignoreIds` field sent with the request.

**Screenshots**
![image](https://user-images.githubusercontent.com/27778301/97784894-296d8180-1b78-11eb-996f-a8965002e34b.png)
![image](https://user-images.githubusercontent.com/27778301/97785041-1dce8a80-1b79-11eb-9f8f-a8b5528dbfe6.png)
![image](https://user-images.githubusercontent.com/27778301/97785115-73a33280-1b79-11eb-8284-a54e1a4c0e6a.png)
![image](https://user-images.githubusercontent.com/27778301/97785136-95041e80-1b79-11eb-99c7-ef04616c58a1.png)
![image](https://user-images.githubusercontent.com/27778301/97785149-a64d2b00-1b79-11eb-90f5-1dfdd2a77f15.png)
![image](https://user-images.githubusercontent.com/27778301/97785168-be24af00-1b79-11eb-91d8-f2950c878b24.png)

ycanardeau:
Thank you!

PR #640, opened 2020-11-01
#639: Fix AlbumApiController.GetTopAlbums

ycanardeau:


VocaDbWeb/Controllers/Api/AlbumApiController.cs
```diff
@@ -249,9 +249,12 @@ public class AlbumApiController : ApiController {
 		[Route("top")]
 		[CacheOutput(ClientTimeSpan = hourInSeconds, ServerTimeSpan = hourInSeconds)]
 		public IEnumerable<AlbumForApiContract> GetTopAlbums(
-			int[] ignoreIds,
+			[FromUri] int[] ignoreIds = null,
 			ContentLanguagePreference languagePreference = ContentLanguagePreference.Default,
-			AlbumOptionalFields fields = AlbumOptionalFields.None) => otherService.GetTopAlbums(languagePreference, fields, ignoreIds ?? new int[0]);
+			AlbumOptionalFields fields = AlbumOptionalFields.None) {
+			ignoreIds ??= Array.Empty<int>();
+			return otherService.GetTopAlbums(languagePreference, fields, Array.Empty<int>()).Where(a => !ignoreIds.Contains(a.Id));
```
ycanardeau, line 256: Since popular albums are cached when the [OtherService.GetTopAlbums](https://github.com/VocaDB/vocadb/blob/846b2b12c07b82cb62054431685fe0bdb9fc6ecd/VocaDbModel/Service/OtherService.cs#L109) method is called, I think this is the right way to filter by IDs. Is it problematic if the cache is invalidated and OtherService.GetTopAlbums is called prior to [OtherService.GetFrontPageContent](https://github.com/VocaDB/vocadb/blob/846b2b12c07b82cb62054431685fe0bdb9fc6ecd/VocaDbModel/Service/OtherService.cs#L450)?

riipah: Then this list won't include those albums that were filtered out by GetFrontPageContent?

I think the best way would be to split GetTopAlbums method so that it can skip caching. This API method is already cached by the CacheOutput attribute, so additional caching is unnecessary. So, when calling from GetFrontPageContent, it would work as before, but when called from this API method, that "service" level cache is ignored. Another fix would be to include ignoreIds list in the cache key, but that might be problematic if the list is long.

ycanardeau: > The problem is still that when calling from GetFrontPageContent, it would not filter out any albums.

Why not?

> I think the best way would be to split GetTopAlbums method so that it can skip ObjectCache caching (do not save anything to cache and do not attempt to retrieve from cache). This API method is already cached by the CacheOutput attribute, so additional caching is unnecessary. So, when calling from GetFrontPageContent, it would work as before, but when called from this API method, that "service" level ObjectCache is ignored.

Do you mean we don't need to [filter out by recent albums](https://github.com/VocaDB/vocadb/blob/846b2b12c07b82cb62054431685fe0bdb9fc6ecd/VocaDbModel/Service/OtherService.cs#L449) when calling from the API? I made [another implementation](https://github.com/VocaDB/vocadb/compare/master...ycanardeau:feat/639-fix-album-api-controller-another-impl?expand=1). If we need to filter out those albums, this implementation won't work.

riipah: > Why not?

I mean, since the list of ignored albums is empty, it gets all albums, caches that and returns that list without filtering, right?

> Do you mean we don't need to [filter out by recent albums](https://github.com/VocaDB/vocadb/blob/846b2b12c07b82cb62054431685fe0bdb9fc6ecd/VocaDbModel/Service/OtherService.cs#L449) when calling from the API? I made [another implementation](https://github.com/VocaDB/vocadb/compare/master...ycanardeau:feat/639-fix-album-api-controller-another-impl?expand=1). If we need to filter out those albums, this implementation won't work.

Of course we should filter out those albums. The link you gave doesn't work for me.

ycanardeau: > I mean, since the list of ignored albums is empty, it gets all albums, caches that and returns that list without filtering, right?

The list of ignored albums is **not** empty when calling from GetFrontPageContent.

riipah: Uh, let me try to explain. Maybe I misunderstood.

1. Someone calls this API method.
2. Top albums list gets cached *without* filtering by recent albums.
3. Someone opens the front page, calling GetFrontPageContent.
4. Top albums list is returned from cache, no filtering is done by recent album IDs (which is wrong).

The above process is what is happening currently, to my understanding.

ycanardeau: Yes, you're right and that's exactly why I asked "Is it problematic if the cache is invalidated and [AlbumApiController.GetTopAlbums](https://github.com/VocaDB/vocadb/blob/846b2b12c07b82cb62054431685fe0bdb9fc6ecd/VocaDbWeb/Controllers/Api/AlbumApiController.cs#L251) is called prior to [OtherService.GetFrontPageContent](https://github.com/VocaDB/vocadb/blob/846b2b12c07b82cb62054431685fe0bdb9fc6ecd/VocaDbModel/Service/OtherService.cs#L450)?". Can we solve this problem by calling GetRecentAlbums from the API method?

riipah: I suggested fixing that by writing that alternate method which skips the cache. Maybe I should just write the code to explain what I mean. It seems pretty clear to me. But I can't do it now.

ycanardeau: > Do you mean we don't need to [filter out by recent albums](https://github.com/VocaDB/vocadb/blob/846b2b12c07b82cb62054431685fe0bdb9fc6ecd/VocaDbModel/Service/OtherService.cs#L449) when calling from the API? I made [another implementation](https://github.com/VocaDB/vocadb/compare/master...ycanardeau:feat/639-fix-album-api-controller-another-impl?expand=1). If we need to filter out those albums, this implementation won't work.

I think I wrote the code as you suggested. Why doesn't the link work for you?

riipah: Ok now I see. Yeah that should work.

>If we need to filter out those albums, this implementation won't work.

Why is that?

riipah: Ok I think I understand what you mean. So when calling from the API, GetTopAlbums won't automatically filter out recent albums, as it does on the front page. It only filters out those albums in the ignoredIds list. That's how I intended it to work. So, calling from the API controller, it only filters out those albums in the ignoredIds. If you want  to filter out recent albums, you need to add them to the ignoredIds list when calling the API.

ycanardeau: > So when calling from the API, GetTopAlbums won't automatically filter out recent albums, as it does on the front page. It only filters out those albums in the ignoredIds list.

That's what I mean.

> So, it seems to me that with the "bool cached" flag it works as expected, although I was originally suggesting having two methods and the one with caching can call the "not cached" version, so you wouldn't need to have any of the caching code in that method.

What would you name those two methods?

> I think the best way would be to split GetTopAlbums method so that it can skip ObjectCache caching (do not save anything to cache and do not attempt to retrieve from cache).

We have another problem with this. Since popular albums are [fetched randomly from the database](https://github.com/VocaDB/vocadb/blob/846b2b12c07b82cb62054431685fe0bdb9fc6ecd/VocaDbModel/Service/OtherService.cs#L91), the results will be inconsistent if we skip the ObjectCache caching.

1. Open the front page. Let's say we have `Album 1` and `Album 2`.
1. Call the API method. `Album 3` and `Album 4`.
1. Open the front page. `Album 1` and `Album 2`. (using ObjectCache)
1. Call the API method. `Album 3` and `Album 4`. (using CacheOutput)
1. Wait for one hour.
1. Open the front page. `Album 1` and `Album 2`. (using ObjectCache)
1. Call the API method. `Album 5` and `Album 6`.

The description says `Gets list of top rated albums, same as front page.`.

riipah: >What would you name those two methods?

I was thinking GetTopAlbumsCached and GetTopAlbums (without cache).

>We have another problem with this. Since popular albums are fetched randomly from the database, the results will be inconsistent if we skip the ObjectCache caching.

I don't see that as problem. What I meant with "same as front page" is that the functionality is the same, but the exact results don't have to be. I would suggest simply changing the description.

VocaDbModel/Service/OtherService.cs
```diff
@@ -98,20 +93,21 @@ class EntryComparer : IEqualityComparer<IEntryWithNames> {
...
-			cache.Add(cacheKey, popularAlbumsCached, DateTime.Now + TimeSpan.FromHours(24));
-
 			return popularAlbumContracts;
 
 		}
 
+		private AlbumForApiContract[] GetTopAlbumsCached(ISession session, int[] recentIds, ContentLanguagePreference languagePreference, AlbumOptionalFields fields) {
+
+			var cacheKey = $"OtherService.PopularAlbums.{languagePreference}";
+			return cache.GetOrInsert(cacheKey, CachePolicy.AbsoluteExpiration(TimeSpan.FromHours(24)), () => GetTopAlbums(session, recentIds, languagePreference, fields));
```
ycanardeau, line 107: Can we use `GetOrInsert` like this?

VocaDbModel/Service/OtherService.cs
```diff
@@ -98,20 +93,21 @@ class EntryComparer : IEqualityComparer<IEntryWithNames> {
 				.OrderByDescending(a => a.RatingAverageInt)
 				.ToArray();
 
-			var popularAlbumsCached = random
-				.Select(a => new TranslatedAlbumContract(a))
```
ycanardeau, line 102: Why was `TranslatedAlbumContract` used here?

VocaDbModel/Service/OtherService.cs
riipah: I don't see why not

VocaDbModel/Service/OtherService.cs
riipah: From the log I can see that languagePreference was added to cache key much later. Originally I intended the same result to be cached for all languages. It could be a mistake that languagePreference was added to cache key. It might be unnecessary. 

It's not a huge issue since it won't cause any invalid results, just that the result is cached for each language separately. And of course using TranslatedAlbumContract is unnecessary if caching per language.

ycanardeau: Should I remove the `TranslatedAlbumContract` class?

ycanardeau:
I think there is no reason to use `TranslatedAlbumContract`.

PR #644, opened 2020-11-19
Add a link to Discord

ycanardeau:


VocaDbWeb/Web.config
```diff
@@ -279,6 +279,7 @@
 			  <link bannerImg="facebook.png" title="Facebook" url="https://www.facebook.com/vocadb"></link>
 			  <link bannerImg="Twitter_Logo.png" title="Twitter" url="https://twitter.com/VocaDB"></link>
 			  <link bannerImg="irc.png" title="IRC" url="https://vocadb.net/Home/Chat"></link>
+			  <link bannerImg="discord-blue.png" title="Discord" url="https://discord.gg/ns39eZ2c"></link>
```
riipah, line 282: This section is actually *not used* by the website. It is replaced during build using configuration transform file. In my opinion, all links configuration should be done in configuration file instead of version control. Updating social media links should not require changes in the version control. Also, these links are different for UtaiteDB and TouhouDB anyway.

ycanardeau: Should I revert these changes?

riipah: It doesn't harm anything, it's just not used. Perhaps it would be best to add XML comment explaining that this socialSites section is configured in the config transformation file. Actually I would recommend moving the whole vocaDb configuration section to that config transformation file.

PR #651, opened 2020-11-24
Replace RenderPartial with Partial

ycanardeau:


VocaDbWeb/Views/Song/Details.cshtml
```diff
@@ -50,7 +50,7 @@
 }
 
 @helper AlbumLink(AlbumContract album) {
-	Html.RenderPartial("Partials/Album/_AlbumLink", new AlbumLinkViewModel(album));
+	@Html.Partial("Partials/Album/_AlbumLink", new AlbumLinkViewModel(album))
```
ycanardeau, line 53: According to [Microsoft Docs](https://docs.microsoft.com/en-us/aspnet/core/mvc/views/partial?view=aspnetcore-5.0), `RenderPartialAsync` doesn't return an IHtmlContent and it streams the rendered output directly to the response. This documentation is for ASP.NET Core, but I think this applies for ASP.NET MVC 5 too. This `AlbumLink` helper is called by [ViewHelper.LinkListHtml](https://github.com/VocaDB/vocadb/blob/d4e98074abf4105f5cfbfcf04cbfb4935d8494bb/VocaDbWeb/Helpers/ViewHelper.cs#L264), which expects `IHtmlString`. However, `RenderPartial` doesn't return `IHtmlString` and it streams the rendered output directly to the response, which is problematic.

riipah: That's correct, that is the difference between RenderPartial and Partial. Clearly LinkListHtml should not be used with RenderPartial. I just wasn't sure why did you change all of them.

ycanardeau: Maybe I shouldn't have changed all of them. I noticed after merging this PR that [RenderPartial provides better performance in some scenarios](https://docs.microsoft.com/en-us/aspnet/core/mvc/views/partial?view=aspnetcore-5.0). But I'm planning to convert all of them to [partial tag helper](https://docs.microsoft.com/en-us/aspnet/core/mvc/views/tag-helpers/built-in/partial-tag-helper?view=aspnetcore-5.0) or [tag helpers](https://docs.microsoft.com/en-us/aspnet/core/mvc/views/tag-helpers/intro?view=aspnetcore-5.0) in ASP.NET Core, so it's not that problem.

Issue #657, opened 2020-11-24
Default sorting events on venue page by date


Nefere256:
**Describe the bug**

If a venue hosts more than one event series, the list of events gets messy as events are sorted by series number. Example: any venue with events of multiple series: [Miyakomesse](https://touhoudb.com/Venue/Details/17), [Nagomix](https://touhoudb.com/Venue/Details/14) or [PiO](https://touhoudb.com/Venue/Details/21).

Usually venue schedules are sorted by date or in calendar form, ex. [nagomix](http://nagomix.co.jp/schedule), [Kyoto Swallow Tail](https://swallowtail.asia/?page_id=16) or [Always Umeda](http://www.always-live.info/Live2020-11.html).

ycanardeau:
Could you please review #658?

PR #658, opened 2020-11-25
Sort events on venue page by date

ycanardeau:
#657 The venues are sorted by date in *ascending* order.

### DIFF
--- a/VocaDbModel/DataContracts/Venues/VenueForApiContract.cs
+++ b/VocaDbModel/DataContracts/Venues/VenueForApiContract.cs
@@ -91,7 +91,7 @@ namespace VocaDb.Model.DataContracts.Venues
 
 			if (fields.HasFlag(VenueOptionalFields.Events))
 			{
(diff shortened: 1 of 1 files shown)


Nefere256 (approved):
It looks okay. 

Thanks for changes!

riipah:
Is this contract used anywhere else such as event series page? Is there any place where we would want to keep the sorting by series number, then date? 

Should the series number sorting still be kept as secondary sort in case date is not specified (the date field is optional)?

ycanardeau:
Thanks for reviews! We don't need to keep the series number as secondary sort because [they're already sorted](https://github.com/VocaDB/vocadb/blob/98ca1ce58e878029cd427f7cd01ee001908bc800/VocaDbModel/Mapping/Venues/VenueMap.cs#L49).

riipah:
As a general design idea, it's better to do sorting like this in the APIs (controller or data contract). The mappings can be changed independently, it's not easy to find out where the sort is used. It also doesn't work for unit tests obviously, because unit tests do not care about NHibernate mappings. If the list is already sorted, then sorting it again will be very fast thanks to quicksort.

ycanardeau:
I see, thanks!

Issue #655, opened 2020-11-24
Display for all comments that a user made


andreoda:
I would like to see all of my comments that I made on VocaDB. The user page says that I made 494 comments so far. It can be useful to check older comments regarding an issue etc.
Maybe a page could open when clicking on "comments: xxx".

riipah:
I'd say #643 is a pre-requirement for this, otherwise it will be complicated.

andreoda:
You mean layout-wise?

riipah:
Technical implementation and performance.

riipah:
It would be good to reuse the page for "recent comments" listing (at /Comments). No need to maintain two separate implementations, and it should be more clear for the user also. The comment author could be just another filter on that page. The layout would be the same also, including grouping comments by entry (to avoid repeating entry information).

Issue #138, opened 2016-04-11
Allow loading more comments in the "recent comments" view


riipah:
Currently limited to 50 comments. It's not possible to load more, because the comments are loaded from multiple tables, so skipping would be expensive.

Two solutions:
- Optimize comment loading so that only IDs are loaded first, then sorted and paginated, finally loading the actual comments, similar to how entries API works.
- Instead of pagination by index, do filtering by comment submission date: it would be fast and easy to load comments that are older than a specific timestamp.

ycanardeau:
Maybe after #643.

Issue #469, opened 2019-02-09
Allow reviewing songs


riipah:
Many songs are not included on albums. It might be useful to be able to post reviews for single songs.

PR #663, opened 2020-11-27
Feat/643 move comments into one table

ycanardeau:
Prepare for #138 and #655.

***

### Current schema:

**AlbumComments** table:
| Name | Data Type | Allow Nulls |
| --- | --- | --- |
| :key: Id | int | |
| Author | int | |
| Created | datetime | |
| Message | nvarchar(4000) | |
| Album | int | |

**ArtistComments** table:
| Name | Data Type | Allow Nulls |
| --- | --- | --- |
| :key: Id | int | |
| Author | int | |
| Created | datetime | |
| Message | nvarchar(4000) | |
| Artist | int | |

**DiscussionComments** table:
| Name | Data Type | Allow Nulls |
| --- | --- | --- |
| :key: Id | int | |
| Author | int | |
| Created | datetime | |
| Message | nvarchar(4000) | |
| Topic | int | |

**DiscussionTopics** table:
| Name | Data Type | Allow Nulls |
| --- | --- | --- |
| :key: Id | int | |
| Content | nvarchar(MAX) | |
| Created | datetime | |
| Deleted | bit | |
| Name | nvarchar(200) | |
| Author | int | |
| Folder | int | |

***

### New schema:

**Comments** table:

| Name | Data Type | Allow Nulls |
| --- | --- | --- |
| :key: Id | int | |
| Author | int | |
| Created | datetime | |
| Deleted | bit | |
| Message | **nvarchar(max)** | |
| LanguageCode | nvarchar(8) | ✔️ |
| Title | nvarchar(200) | ✔️ |
| EntryType | varchar(20) | |
| Album | int | ✔️ |
| Artist | int | ✔️ |
| DiscussionTopic | int | ✔️ |
| ReleaseEvent | int | ✔️ |
| Song | int | ✔️ |
| SongList | int | ✔️ |
| Tag | int | ✔️ |
| User | int | ✔️ |

**DiscussionTopics** table:
| Name | Data Type | Allow Nulls |
| --- | --- | --- |
| :key: Id | int | |
| Created | datetime | |
| Deleted | bit | |
| Name | nvarchar(200) | |
| Folder | int | |

Raw SQL query on the front page (comments):
```
select
	comment0_.Id as id1_25_,
	comment0_.[Created] as column3_25_,
	comment0_.[Deleted] as column4_25_,
	comment0_.[Message] as column5_25_,
	comment0_.[Author] as column6_25_,
	comment0_.[Album] as column7_25_,
	comment0_.[LanguageCode] as column8_25_,
	comment0_.[Title] as column9_25_,
	comment0_.[Artist] as column10_25_,
	comment0_.[Topic] as column11_25_,
	comment0_.[ReleaseEvent] as column12_25_,
	comment0_.[Song] as column13_25_,
	comment0_.SongList as songlist14_25_,
	comment0_.[Tag] as column15_25_,
	comment0_.[User] as column16_25_,
	comment0_.[EntryType] as column2_25_
from dbo.Comments comment0_
where  not (comment0_.[Deleted]=1) and  not (comment0_.[EntryType]='UserComment')
order by comment0_.[Created] desc
OFFSET 0 ROWS FETCH FIRST ? ROWS ONLY
```

Raw SQL query on the front page (activity entries):
```
select
	activityen0_.Id as id1_0_,
	activityen0_.[CreateDate] as column3_0_,
	activityen0_.[EditEvent] as column4_0_,
	activityen0_.[Author] as column5_0_,
	activityen0_.[ArchivedAlbumVersion] as column6_0_,
	activityen0_.[Album] as column7_0_,
	activityen0_.[ArchivedArtistVersion] as column8_0_,
	activityen0_.[Artist] as column9_0_,
	activityen0_.[ArchivedReleaseEventVersion] as column10_0_,
	activityen0_.[ReleaseEvent] as column11_0_,
	activityen0_.[ArchivedSongVersion] as column12_0_,
	activityen0_.[Song] as column13_0_,
	activityen0_.[ArchivedSongListVersion] as column14_0_,
	activityen0_.[SongList] as column15_0_,
	activityen0_.[ArchivedTagVersion] as column16_0_,
	activityen0_.[Tag] as column17_0_,
	activityen0_.[ArchivedVenueVersion] as column18_0_,
	activityen0_.[Venue] as column19_0_,
	activityen0_.[EntryType] as column2_0_
from dbo.ActivityEntries activityen0_
order by activityen0_.[CreateDate] desc
OFFSET 0 ROWS FETCH FIRST ? ROWS ONLY 
```


VocaDb.Migrations/Migrations.cs
```diff
@@ -1,10 +1,95 @@
...
+				.WithColumn("Id").AsInt64().NotNullable().Identity().PrimaryKey()
+				.WithColumn("Author").AsInt32().NotNullable().ForeignKey(TableNames.Users, "Id").OnDelete(Rule.Cascade)
+				.WithColumn("Created").AsDateTime().NotNullable()
+				.WithColumn("Deleted").AsBoolean().NotNullable().WithDefaultValue(false)
+				.WithColumn("Message").AsString(4000).NotNullable()
+				// TODO: remove these columns
+				.WithColumn("OldTable").AsString().Nullable()
+				.WithColumn("OldId").AsInt32().Nullable();
+
+			var values = string.Join(" union ", commentTables.Select(table => $"select '{table.NameWithSchema}', Id, Author, Created, Message from {table.NameWithSchema}"));
```
riipah, line 59: Maybe add a comment here with example of what kind of SQL is generated.

VocaDb.Migrations/Migrations.cs
```diff
@@ -1,10 +1,95 @@
...
+			Execute.Sql($@"insert into Comments(OldTable, OldId, Author, Created, Message) {values} order by Created");
+
+			foreach (var table in commentTables)
+			{
+				Create.Column("Comment").OnTable(table.Name).InSchema(table.Schema).AsInt64().Nullable();
```
riipah, line 64: Should this be marked a primary key? And unique most likely. I know you're still editing this...

riipah:
I see you decided to go with the table per type strategy after all. I guess that's fine. The main possible issue is performance, but probably that won't be a problem unless you're loading hundreds of comments at a time.

ycanardeau:
[I was initially trying to use Table-per-hierarchy (TPH) inheritance strategy](https://github.com/VocaDB/vocadb/pull/663/commits/37c630a30596da549e6fbe52cb03aacd3b619a64), but it turned out to be complicated. How would you distinguish `AlbumComments` and `AlbumReviews`? For example, we could add a new enum like `CommentType`?

riipah:
To be honest, I'm thinking it could be better to keep reviews separate, because reviews are not exactly comments, even though they are sometimes shown in the same listing. But when browsing user's comments for example, we can skip reviews I think. So including reviews with comments is a special situation, and since it's only one type, it can be handled the way that it's done currently, by cross joining the two listings when needed.

ycanardeau:
It's possible to implement new review types in the future, isn't it? #469

riipah:
True, and in that case, I would make a base class for reviews, but it's a separate hierarchy from comments. Comments and reviews are similar, but not the same.

riipah:
And this table per type mapping may be fine too for comments. At least it's the "purest" from database point of view, because you don't have null columns. You should need to test the performance, so that comment listing doesn't become too slow because of the joins. All the inheritance strategies have their pros and cons, and it depends on the situation which one to choose. Table per hierarchy (all in 1 table) has the best performance, but it's not the cleanest database schema.

ycanardeau:
And I'm considering something like this:

**DiscussionTopics** table:

| Name | Data Type | Allow Nulls |
| --- | --- | --- |
| FirstComment | **bigint** | |
| Locked | bit | |
| Pinned | bit | |
| Name | nvarchar(200) | |
| Folder | int | |

**DiscussionComments** table:

| Name | Data Type | Allow Nulls |
| --- | --- | --- |
| Comment | **bigint** | |
| Topic | int | |

That's why we have to change `nvarchar(4000)` to `nvarchar(max)`.

riipah:
Implementing a search index like ElasticSearch would be good for the search page too, where artists, albums and songs are cross joined. With ElasticSearch you could keep the master data in SQL and have additional listings for scenarios such as search page.

riipah:
> That's why we have to change `nvarchar(4000)` to `nvarchar(max)`

Yes that's fine.

ycanardeau:
> Yes that's fine.

What about the schema?

ycanardeau:
> You only need to test the performance, so that comment listing doesn't become too slow because of the joins.

Yes, I'm going to see the raw SQL queries.

riipah:
> What about the schema? A discussion topic can be seen as a comment in this way.

Originally I intended the discussions to be separate from the rest of the website. They are not "database content". That's why I even made them a custom schema. So IMO discussion comments could still be kept in their own table as they are currently, separate from the rest of the comments, treating them a special case like reviews. But it's also ok to bring them to the same table with other comments, or joining them to the base comments table I guess.

VocaDbModel/Mapping/CommentMap.cs
```diff
@@ -1,20 +1,124 @@
...
 	{
 		protected CommentMap()
 		{
+			Table("Comments");
 			Cache.ReadWrite();
 			Id(m => m.Id);
 
 			Map(m => m.Created).Not.Nullable();
-			Map(m => m.Message).Length(4000).Not.Nullable();
+			Map(m => m.Deleted).Not.Nullable();
```
ycanardeau, line 15: Added the `Deleted` column to implement soft deletes for comments.

VocaDbModel/Mapping/CommentMap.cs
```diff
@@ -1,20 +1,124 @@
...
+			Map(m => m.Message).Length(int.MaxValue).Not.Nullable();
```
ycanardeau, line 16: Changed max length from `4000` to `int.MaxValue`.

VocaDb.Migrations/Migrations.cs
```diff
@@ -1,10 +1,149 @@
...
+
+			// Rename columns on `AlbumReviews` table.
+			Rename.Column("Date").OnTable(TableNames.AlbumReviews).To("Created");
+			Rename.Column("Text").OnTable(TableNames.AlbumReviews).To("Message");
+			Rename.Column("User").OnTable(TableNames.AlbumReviews).To("Author");
+			Delete.Index("UX_AlbumReviews").OnTable(TableNames.AlbumReviews);
+
+			// Create `Comments` table.
+			Create.Table(TableNames.Comments)
+				.WithColumn("Id").AsInt32().NotNullable().Identity().PrimaryKey()
```
ycanardeau, line 55: Let's use `int` rather than `long` for now.

VocaDb.Migrations/Migrations.cs
```diff
@@ -1,10 +1,149 @@
...
+			// Make `Comment` nullable and rename it to `OldComment`.
+			Alter.Column("Comment").OnTable(TableNames.DiscussionTopics).InSchema(SchemaNames.Discussions).AsInt32().Nullable();
+			Rename.Column("Comment").OnTable(TableNames.DiscussionTopics).InSchema(SchemaNames.Discussions).To("OldComment");
+
+			Rename.Column("OldCreated").OnTable(TableNames.DiscussionTopics).InSchema(SchemaNames.Discussions).To("Created");
+		}
+
+		public override void Down()
+		{
+			throw new System.NotImplementedException();
```
ycanardeau, line 143: Not implemented/supported.

VocaDb.Migrations/Migrations.cs
```diff
@@ -1,10 +1,149 @@
...
+
+			// insert into Comments(OldTable, OldId, Author, Created, Message)
+			//     select 'AlbumComments', Id, Author, Created, Message from AlbumComments union
+			//     select 'ArtistComments', Id, Author, Created, Message from ArtistComments union
+			//     select 'discussions.DiscussionComments', Id, Author, Created, Message from discussions.DiscussionComments union
+			//     ...
+			//     select 'UserComments', Id, Author, Created, Message from UserComments union
+			//     select 'DiscussionTopics', Id, Author, Created, Message from discussions.DiscussionTopics union
+			//     select 'AlbumReviews', Id, Author, Created, Message from AlbumReviews
+			// order by Created
```
ycanardeau, line 87: Is it enough just to sort by creation date?

VocaDb.Migrations/Migrations.cs
```diff
@@ -1,10 +1,149 @@
...
+				// Make `Message` nullable and rename it to `OldMessage`.
+				Alter.Column("Message").OnTable(table.Name).InSchema(table.Schema).AsString(4000).Nullable();
+				Rename.Column("Message").OnTable(table.Name).InSchema(table.Schema).To("OldMessage");
+
+				if (Schema.Schema(table.Schema).Table(table.Name).Column("AuthorName").Exists())
+				{
+					// Make `AuthorName` nullable and rename it to `OldAuthorName`.
+					Alter.Column("AuthorName").OnTable(table.Name).InSchema(table.Schema).AsString(100).Nullable();
+					Rename.Column("AuthorName").OnTable(table.Name).InSchema(table.Schema).To("OldAuthorName");
+				}
```
ycanardeau, line 118: Why do `AlbumComments` and `DiscussionComments` tables have the `AuthorName` column? Can we remove it?

VocaDbModel/Domain/Discussions/DiscussionTopic.cs
```diff
@@ -60,22 +59,18 @@ public virtual IList<DiscussionComment> Comments
...
-		public virtual string Content
-		{
-			get => content;
-			set
-			{
-				ParamIs.NotNullOrEmpty(() => value);
-				content = value;
-			}
-		}
+		public virtual IEnumerable<DiscussionComment> Comments => AllComments.Where(c => c != FirstComment).Where(c => !c.Deleted);
```
ycanardeau, line 62: The `FirstComment` is regarded as the content of a topic, therefore, we exclude it in `Comments`.

VocaDbModel/Service/OtherService.cs
```diff
@@ -275,48 +264,16 @@ private IEnumerable<EntryWithCommentsContract> CreateEntryWithCommentsContract<T
...
-			return await q.OrderByDescending(c => c.Created).Take(maxComments).Cast<Comment>().VdbToListAsync();
-		}
-
 		private async Task<EntryWithCommentsContract[]> GetRecentCommentsAsync(ISession session, int maxComments)
 		{
-			var albumComments = await GetComments<Album, AlbumComment>(session, maxComments, true);
-			var artistComments = await GetComments<Artist, ArtistComment>(session, maxComments, true);
-			var songComments = await GetComments<Song, SongComment>(session, maxComments, true);
-			var discussionComments = await GetComments<DiscussionTopic, DiscussionComment>(session, maxComments, true);
-			var songListComments = await GetComments<SongList, SongListComment>(session, maxComments, false);
```
ycanardeau, line 294: Why was `checkDeleted` `false` here?

VocaDbModel/Service/OtherService.cs
```diff
@@ -170,16 +169,6 @@ private ReleaseEventForApiContract[] GetRecentEvents(ISession session)
...
-		private Task<EntryWithCommentsContract[]> GetRecentCommentsAsync(ISession session)
-		{
-			var cacheKey = $"OtherService.RecentComments.{LanguagePreference}";
-			return cache.GetOrInsert(cacheKey, CachePolicy.AbsoluteExpiration(TimeSpan.FromMinutes(5)), async () =>
-			{
-				var item = await GetRecentCommentsAsync(session, 9);
-				return item;
-			});
-		}
-
```
ycanardeau, line 182: Disabled cache.

VocaDbModel/Database/Queries/UserQueries.cs
```diff
@@ -191,13 +192,8 @@ private CachedUserStats GetCachedUserStats(IDatabaseContext<User> ctx, User user
 					stats.ArtistCount = GetArtistCount(ctx, user);
 					stats.FavoriteSongCount = GetSongCount(ctx, user);
 
-					stats.CommentCount
-						= ctx.Query<AlbumComment>().Count(c => c.Author.Id == user.Id)
-						+ ctx.Query<ArtistComment>().Count(c => c.Author.Id == user.Id)
-						+ ctx.Query<SongComment>().Count(c => c.Author.Id == user.Id);
```
ycanardeau, line 197: I suspect the original implementation was wrong.

VocaDbModel/Service/OtherService.cs
```diff
@@ -275,48 +264,16 @@ private IEnumerable<EntryWithCommentsContract> CreateEntryWithCommentsContract<T
...
 		private async Task<EntryWithCommentsContract[]> GetRecentCommentsAsync(ISession session, int maxComments)
 		{
-			var albumComments = await GetComments<Album, AlbumComment>(session, maxComments, true);
-			var artistComments = await GetComments<Artist, ArtistComment>(session, maxComments, true);
-			var songComments = await GetComments<Song, SongComment>(session, maxComments, true);
-			var discussionComments = await GetComments<DiscussionTopic, DiscussionComment>(session, maxComments, true);
-			var songListComments = await GetComments<SongList, SongListComment>(session, maxComments, false);
-			var tagComments = await GetComments<Tag, TagComment>(session, maxComments, true);
-			var eventComments = await GetComments<ReleaseEvent, ReleaseEventComment>(session, maxComments, true);
-
```
ycanardeau, line 297: Is there any reason not to include `UserComment`s here?

VocaDb.Migrations/Migrations.cs
```diff
@@ -1,10 +1,149 @@
...
+		public override void Up()
+		{
+			// Rename columns on `DiscussionTopics` table.
+			Rename.Column("Content").OnTable(TableNames.DiscussionTopics).InSchema(SchemaNames.Discussions).To("Message");
+
+			// Rename columns on `AlbumReviews` table.
+			Rename.Column("Date").OnTable(TableNames.AlbumReviews).To("Created");
+			Rename.Column("Text").OnTable(TableNames.AlbumReviews).To("Message");
+			Rename.Column("User").OnTable(TableNames.AlbumReviews).To("Author");
+			Delete.Index("UX_AlbumReviews").OnTable(TableNames.AlbumReviews);
```
ycanardeau, line 51: Deleted `UX_AlbumReviews` to implement soft deletes for reviews.

VocaDbModel/Service/OtherService.cs
riipah: Perhaps user page comments can be considered more personal, no need to show them on the front page.

VocaDb.Migrations/Migrations.cs
riipah: Perhaps you could add it as normal non-unique index instead?

VocaDbModel/Database/Queries/UserQueries.cs
riipah: If you're referring it to not counting all entry types, I probably just didn't update it when adding more comment types. It's only informational anyway.

VocaDb.Migrations/Migrations.cs
```diff
@@ -1,10 +1,149 @@
...
+				}
+
+				if (table.EntryTypeName != null)
+				{
+					Delete.PrimaryKey($"PK_{table.Name}").FromTable(table.Name).InSchema(table.Schema);
+					// Make `Id` nullable and rename it to `OldId`.
+					Alter.Column("Id").OnTable(table.Name).InSchema(table.Schema).AsInt32().Nullable();
+					Rename.Column("Id").OnTable(table.Name).InSchema(table.Schema).To("OldId");
+
+					Create.PrimaryKey().OnTable(table.Name).WithSchema(table.Schema).Columns("Comment", table.EntryTypeName);
```
ycanardeau, line 127: Is it okay to use a composite key?

VocaDb.Migrations/Migrations.cs
riipah: Originally there was an idea that you could delete the user account (remove it from database), making the Author field NULL, but preserving the name of the user who posted the comment. But I scrapped that idea at some point, it's better to just keep user references non-nullable and clear user information if needed, never deleting the user entity.

I guess there is still some value keeping record of the name that user had when posting the comment (users can change their username), but I don't see it as necessary.

VocaDb.Migrations/Migrations.cs
riipah: Comments are in thousands anyway, so int32 is fine.

VocaDb.Migrations/Migrations.cs
riipah: Should be.

VocaDb.Migrations/Migrations.cs
riipah: I think composite key is unnecessary. The Comment column by itself is already a unique key. I don't know if it makes difference to NHibernate, but it shouldn't.

VocaDb.Migrations/Migrations.cs
```diff
@@ -1,10 +1,149 @@
...
+				}
+			}
+
+			Execute.Sql($"update Comments set Deleted = ec.Deleted from discussions.DiscussionTopics ec inner join Comments c on c.OldTable = 'discussions.DiscussionTopics' and ec.Id = c.OldId");
+			Execute.Sql($"insert into discussions.DiscussionComments(Comment, Topic) select Comment, Id from discussions.DiscussionTopics");
```
riipah, line 132: Could you comment this?

VocaDbModel/Domain/Discussions/DiscussionTopic.cs
riipah: Please add a comment that mentions this, so we will remember it even merging the PR.

VocaDbModel/Database/Queries/AlbumQueries.cs
```diff
@@ -177,33 +178,34 @@ public AlbumReviewContract AddReview(int albumId, AlbumReviewContract contract)
...
 				}
 				else
 				{
-					review = ctx.Query<AlbumReview>().FirstOrDefault(r => r.Album.Id == albumId && r.User.Id == PermissionContext.LoggedUserId && r.LanguageCode == contract.LanguageCode);
+					review = ctx.Query<AlbumReview>().FirstOrDefault(r => r.EntryForComment.Id == albumId && r.Author.Id == PermissionContext.LoggedUserId && r.LanguageCode == contract.LanguageCode);
 				}
 
 				// Create
-				if (review == null)
+				if (review == null || review.Deleted)
```
riipah, line 192: What happens if the review is deleted and a new review is posted? Won't this still return the deleted review even after that and then make a new review instead of updating?

1. User posts review (review is created)
2. User deletes review (review is marked deleted)
3. User posts new review (review is created)
4. User updates review (???)

VocaDbModel/Database/Queries/AlbumQueries.cs
```diff
@@ -363,9 +365,10 @@ public AlbumDetailsContract GetAlbumDetails(int id, string hostname)
 
 				var stats = session.Query<Album>()
```
riipah, line 366: You should move the comment loading and hit counting to separate calls.

The reason it's done like this is to load both of these stats using a single call, but because of the ToArray() you added, it's loading them separately anyway, so it's unnecessary. Best would be to cache these values also.

riipah:
There is quite a lot of code here, so it's going to take several days to review. Just letting you know, to be patient.

VocaDbModel/Service/OtherService.cs
riipah: Soft-deletion was added to song lists afterwards

riipah:
Could you add DiscussionTopic and DiscussionComment tables to the "new schema", because it's a bit unclear to me.

VocaDbModel/Domain/Discussions/DiscussionTopic.cs
```diff
@@ -108,10 +103,12 @@ public virtual string Name
 		public virtual Comment CreateComment(string message, AgentLoginData loginData)
 		{
 			var comment = new DiscussionComment(this, message, loginData);
-			Comments.Add(comment);
+			AllComments.Add(comment);
 			return comment;
 		}
 
+		public virtual void Delete() => Deleted = true;
```
riipah, line 110: What happens to DiscussionComments when DiscussionTopic is deleted?

VocaDbModel/Service/OtherService.cs
```diff
@@ -275,48 +264,16 @@ private IEnumerable<EntryWithCommentsContract> CreateEntryWithCommentsContract<T
...
-			{
-				Created = t.Created
-			});
-
-			var albumReviews = await session.Query<AlbumReview>().OrderByDescending(r => r.Date).Take(maxComments).VdbToListAsync();
-			var albumReviewsAsComments = albumReviews.Select(r => new AlbumComment(r.Album, r.Text, new AgentLoginData(r.User))
-			{
-				Created = r.Date
-			});
+			var comments = (await session.Query<Comment>()
```
riipah, line 269: Since you're doing the ```.Where(c => !c.Entry.Deleted);``` filtering last after ToList, isn't it going to return less comments if there are deleted entries?

VocaDbModel/Database/Queries/AlbumQueries.cs
ycanardeau: Users can successfully update the new review. The `else` block gets called if and only if `review != null && !review.Deleted` (De Morgan's laws).

VocaDbModel/Service/OtherService.cs
ycanardeau: Yes, but this happens in activity entries as well. I don't think we can directly query `Entry.Deleted` before `ToList`. (an ORM limitation?) I think we could have a column something like `EntryDeleted` as a workaround.

ycanardeau:
> Could you add DiscussionTopic and DiscussionComment tables to the "new schema", because it's a bit unclear to me.

The `DiscussionTopics` table doesn't have `Content` (`Message`) and `Author` anymore. Instead, the `DiscussionComments` table does.

riipah:
I tried to test this locally but got

```
!!! An error occured executing the following sql:
ALTER TABLE [discussions].[DiscussionTopics] ALTER COLUMN [Message] NVARCHAR(4000)
The error was String or binary data would be truncated.
The statement has been terminated.
```

Should it be
```csharp
Alter.Column("Message").OnTable(table.Name).InSchema(table.Schema).AsString(int.MaxValue).Nullable();
```

VocaDbModel/Database/Queries/AlbumQueries.cs
riipah: I didn't notice the if block before. So when updating a review, the ID is always non-null and the review gets loaded by ID. Only when creating a new review it's using the query (```ctx.Query<AlbumReview>().FirstOrDefault...```). I was thinking it would use the query every time, which wouldn't work for deleted reviews.

Could you add unit tests for this, since it became more complicated now?

riipah: It's also ok to have a separate issue for the tests, but I think it should be tested.

VocaDbModel/Service/OtherService.cs
riipah: Right. I didn't notice Entry is an accessor property (it's hard when doing review on GitHub). ORM can't map those directly. Maybe there is a workaround since this is a common situation (accessing polymorphic property). For now it's ok like this too, but maybe add a comment?

VocaDb.Migrations/Migrations.cs
```diff
@@ -1,10 +1,157 @@
...
+			/// <summary>
+			/// Entry type name. (e.g. `Album`, `Artist`, `Topic`, ...)
+			/// </summary>
+			public string EntryTypeName { get; }
+
+			/// <summary>
+			/// Table name with schema. (e.g. `AlbumComments`, `ArtistComments`, `discussions.DiscussionComments`, ...)
+			/// </summary>
+			public string NameWithSchema => Schema != null ? $"{Schema}.{Name}" : Name;
+		}
```
dset0x, line 40: I might be too late now, but wouldn't it make more sense to have the above strings be some sort of enumerations (if available), or is type safety ensured elsewhere? There may also be a performance benefit, but then again I suppose indexing takes care of that.

It also looks odd to me to have information duplication in `NameWithSchema`. Shouldn't this filtering be encapsulated in logic, or is this for performance reasons?

(I have no clue about these frameworks, or many best practices, so apologies if my comments seem more like incoherent rambling)

VocaDb.Migrations/Migrations.cs
```diff
@@ -1,10 +1,157 @@
...
+			//     select 'AlbumComments', Id, Author, Created, Message from AlbumComments union
+			//     select 'ArtistComments', Id, Author, Created, Message from ArtistComments union
+			//     select 'discussions.DiscussionComments', Id, Author, Created, Message from discussions.DiscussionComments union
+			//     ...
+			//     select 'UserComments', Id, Author, Created, Message from UserComments union
+			//     select 'DiscussionTopics', Id, Author, Created, Message from discussions.DiscussionTopics union
+			//     select 'AlbumReviews', Id, Author, Created, Message from AlbumReviews
+			// order by Created
+			var values = string.Join(" union ", commentTables.Select(table => $"select '{table.NameWithSchema}', Id, Author, Created, Message from {table.NameWithSchema}"));
+			Execute.Sql($@"insert into {TableNames.Comments}(OldTable, OldId, Author, Created, Message) {values} order by Created");
```
dset0x, line 93: I am perplexed and concerned that there is a need to run SQL. Is this how this normally works, or is there a special need here?

riipah: For migrations, unless it's a basic update or insert, you may need to write SQL. As far as I know, there is no way to do update like this using FluentMigrator without SQL. But it's ok, sometimes SQL is the easiest.

VocaDb.Migrations/Migrations.cs
ycanardeau: The `CommentTable` is for one-time use only. We have 10 different comment tables: `AlbumComments`, `ArtistComments`, 
`discussions.DiscussionComments`, `ReleaseEventComments`, `SongComments`, `SongListComments`, `TagComments`, `UserComments`, `discussions.DiscussionTopics` and `AlbumReviews`. As you can see, `discussions.DiscussionComments` and `discussions.DiscussionTopics` reside in a different schema. So I think this is the easiest way.

VocaDb.Migrations/Migrations.cs
```diff
@@ -41,18 +40,19 @@ public CommentTable(string name, string schema = null, string entryTypeName = nu
...
 
 			// Rename columns on `AlbumReviews` table.
 			Rename.Column("Date").OnTable(TableNames.AlbumReviews).To("Created");
 			Rename.Column("Text").OnTable(TableNames.AlbumReviews).To("Message");
 			Rename.Column("User").OnTable(TableNames.AlbumReviews).To("Author");
-			Delete.Index("UX_AlbumReviews").OnTable(TableNames.AlbumReviews);
-			Create.Index().OnTable(TableNames.AlbumReviews)
-				.OnColumn("Album").Ascending()
-				.OnColumn("Author").Ascending()
-				.OnColumn("LanguageCode").Ascending();
```
ycanardeau: Should I still add an index on the `Comments` table? Should it be `Album, Author, LanguageCode` or `Author, LanguageCode`?

ycanardeau (commented):
Since TPT was too complicated, I converted it to TPH.

VocaDb.Migrations/Migrations.cs
riipah: As a general rule, usually you want to add indexes to reference columns such as Album/Artist/Song, and also Author, because joins between database tables are more efficient with indexes. A separate index for each reference column of course, because they are generally not used together. You can add other columns, if those columns are generally used together with the reference columns.

VocaDbModel/Service/QueryableExtensions/CommentQueryableExtensions.cs
```diff
@@ -0,0 +1,13 @@
+using System.Linq;
+using VocaDb.Model.Domain;
+using VocaDb.Model.Domain.Comments;
+
+namespace VocaDb.Model.Service.QueryableExtensions
+{
+	public static class CommentQueryableExtensions
```
riipah, line 7: Could you just make Comment implement IDeletableEntry?

riipah:
Thanks again, it looks ok now. Needs to be tested properly on the beta site though, because it's a rather big database schema change.

VocaDbModel/Service/QueryableExtensions/CommentQueryableExtensions.cs
ycanardeau: Originally I intended the comment ID to be int64, so I didn't make `Comment` implement `IDeletableEntry`.

ycanardeau:
@dset0x @riipah Deployed on the beta site. Thank you for your reviews!

PR #664, opened 2020-11-28
Update Web.config

ycanardeau:
The original exception:
`FileLoadException Could not load file or assembly 'System.Runtime.CompilerServices.Unsafe, Version=4.0.4.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies. The located assembly's manifest definition does not match the assembly reference. (Exception from HRESULT: 0x80131040) Void LoadManagedCodePage()`

Note: `VocaDb.BandcampMetadataExtractor` depends on `AngleSharp`, which depends on `System.Text.Encoding.CodePages`, which depends on `System.Runtime.CompilerServices.Unsafe`.

References:
[Could not load file or assembly System.Runtime.CompilerServices.Unsafe](https://stackoverflow.com/questions/62764744/could-not-load-file-or-assembly-system-runtime-compilerservices-unsafe)
[Binding Redirects](https://nickcraver.com/blog/2020/02/11/binding-redirects/)

### DIFF
--- a/VocaDbWeb/Web.config
+++ b/VocaDbWeb/Web.config
@@ -43,6 +43,7 @@
       <assemblies>
         <!-- http://www.lyalin.com/2014/04/25/the-type-system-object-is-defined-in-an-assembly-that-is-not-reference-mvc-pcl-issue/ -->
         <add assembly="System.Runtime, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a" />
+        <add assembly="System.Runtime.CompilerServices.Unsafe, Version=4.0.4.1, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a" />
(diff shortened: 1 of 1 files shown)


VocaDbWeb/Web.config
```diff
@@ -43,6 +43,7 @@
       <assemblies>
         <!-- http://www.lyalin.com/2014/04/25/the-type-system-object-is-defined-in-an-assembly-that-is-not-reference-mvc-pcl-issue/ -->
         <add assembly="System.Runtime, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a" />
+        <add assembly="System.Runtime.CompilerServices.Unsafe, Version=4.0.4.1, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a" />
```
riipah, line 46: I think this should not be necessary

ycanardeau: You're right. Thanks.

Issue #630, opened 2020-10-16
GoogleBot can't find the lyrics


ycanardeau:
Compare `あれ、なんだかアタシの名前みたい（笑）` with `site:vocadb.net あれ、なんだかアタシの名前みたい（笑）` on Google Search.

riipah:
https://developers.google.com/search/docs/guides/fix-search-javascript

It sounds like adding history API URLs (client-side routing) to the tabs on song pages could fix the problem.

ycanardeau:
I think we can use the history API for the search page as well. For example, if I search a song, click it and go back to the search page, then the search result will be lost.

riipah:
Ideally yes, but for the search page it's going to be much more complicated because there is so many options. For song/artist/album pages it's much easier.

ycanardeau:
What should the URL look like? For example, `https://vocadb.net/S/186287/Lyrics/54513`?

riipah:
Something like that yes, maybe lowercase "l" and maybe you could append the language too. So, ```https://vocadb.net/S/186287/lyrics/54513/english```

ycanardeau:
I noticed that comments in the Discussion tab on the song details page are indexed by Google. And I think I found the cause of this issue. If you access the following URL directly, the lyrics won't appear. https://vocadb.net/S/186287#lyricsTab

riipah:
Right. Well, fixing that could be a quick fix, so that the content loads when navigating to URL directly.

But I think that overall, routing to tabs by URL fragment (#) is not so recommended anymore. Using the history API with proper client side routing (/S/3939/lyrics...) would still be better in the long run.

PR #681, opened 2020-12-06
Render lyrics on the server side

ycanardeau:
Workaround for #630.

### DIFF
--- a/VocaDbWeb/Views/Song/Details.cshtml
+++ b/VocaDbWeb/Views/Song/Details.cshtml
@@ -466,16 +466,28 @@
 					</li>
 				}
 			</ul>
-			
-			<div data-bind="with: selectedLyrics">
-				<p data-bind="visible: source || url">
-					(@Res.LyricsFrom
-					<span data-bind="visible: url"><a class="extLink" data-bind="text: source || url, attr: { href: url, onclick: 'app.functions.trackOutboundLink(event)' }"></a>)</span>					
-					<span data-bind="visible: !url, text: source + ')'"></span>
-				</p>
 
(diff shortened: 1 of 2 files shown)


VocaDbWeb/Views/Song/Details.cshtml
```diff
@@ -466,16 +466,28 @@
...
+								<span><a class="extLink" href="@lyrics.URL" onclick="app.functions.trackOutboundLink(event)">@(!string.IsNullOrEmpty(lyrics.Source) ? lyrics.Source : lyrics.URL)</a>)</span>
+							}
+							else
+							{
+								<span><text>@lyrics.Source</text>)</span>
+							}
+						</p>
+					}
+
+					<p class="song-lyrics">@lyrics.Value</p>
```
riipah, line 488: But now you're including all the lyrics content on the song page, which will potentially add a significant amount of data. Is it worth it just for Google? Did you compare the song page load time?

ycanardeau: I've reverted the changes.

dset0x: Did you find out if it actually causing performance issues?

riipah: Even if the difference is just 10ms, you know that if the song has lyrics in multiple languages, it can be hundreds of kb extra transfer. Better to load that lazily. Google counts both data transfer amount and speed. It's not an important issue anyway, and there is a way to make the lyrics appear in Google while keeping the lazy load, which in my opinion is the best way.

PR #700, opened 2020-12-21
Enable unsafe binary serialization

ycanardeau:
#598

### DIFF
--- a/VocaDbWeb.Core/VocaDbWeb.Core.csproj
+++ b/VocaDbWeb.Core/VocaDbWeb.Core.csproj
@@ -5,6 +5,8 @@
     <RootNamespace>VocaDb.Web</RootNamespace>
     <LangVersion>preview</LangVersion>
     <Nullable>enable</Nullable>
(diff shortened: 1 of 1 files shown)


VocaDbWeb.Core/VocaDbWeb.Core.csproj
```diff
@@ -5,6 +5,8 @@
     <RootNamespace>VocaDb.Web</RootNamespace>
     <LangVersion>preview</LangVersion>
     <Nullable>enable</Nullable>
+    <!-- CA2300: Do not use insecure deserializer BinaryFormatter -->
+    <EnableUnsafeBinaryFormatterSerialization>true</EnableUnsafeBinaryFormatterSerialization>
```
ycanardeau, line 9: Can we enable unsafe binary serialization temporally? Or should we replace it with a safe one *now*?

riipah: It's ok to have this for now, but maybe add to the comment mentioning where is it being used, and note that it should be removed when possible.

PR #701, opened 2020-12-23
Copy `Controllers` from VocaDbWeb to VocaDbWeb.Core

ycanardeau:
#598


VocaDbWeb.Core/Controllers/Api/AlbumApiController.cs
```diff
@@ -0,0 +1,338 @@
...
+		/// Optional fields (optional). Possible values are artists, names, pvs, tags, tracks, webLinks.
+		/// </param>
+		/// <param name="lang">Content language preference (optional).</param>
+		/// <returns>Page of albums.</returns>
+		/// <example>https://vocadb.net/api/albums?query=Synthesis&amp;discTypes=Album</example>
+		[HttpGet("")]
+		public PartialFindResult<AlbumForApiContract> GetList(
+			string query = "",
+			DiscType discTypes = DiscType.Unknown,
+			[FromQuery(Name = "tagName[]")] string[] tagName = null,
```
ycanardeau, line 161: Replaced `FromUri` with `FromQuery` plus `Name`.

VocaDbWeb.Core/Controllers/Api/ResourcesApiController.cs
```diff
@@ -0,0 +1,105 @@
...
+		// TODO: implement
+		/*class DefaultCasingConfig : Attribute, IControllerConfiguration
+		{
+			public void Initialize(HttpControllerSettings controllerSettings, HttpControllerDescriptor controllerDescriptor)
+			{
+				controllerSettings.Formatters.Clear();
+				controllerSettings.Formatters.Add(new JsonMediaTypeFormatter());
+				controllerSettings.Formatters.JsonFormatter.SerializerSettings.ContractResolver = new DefaultContractResolver();
+			}
+		}*/
```
ycanardeau, line 32: Why do we need this?

VocaDbWeb.Core/Controllers/HomeController.cs
```diff
@@ -0,0 +1,105 @@
...
+using VocaDb.Model.Database.Queries;
+using VocaDb.Model.Domain;
+using VocaDb.Model.Service;
+using VocaDb.Model.Service.BrandableStrings;
+using VocaDb.Model.Service.Search;
+using VocaDb.Web.Models;
+
+namespace VocaDb.Web.Controllers
+{
+	// TODO: implement [SessionState(SessionStateBehavior.ReadOnly)]
```
ycanardeau, line 15: How can we achieve this in ASP.NET Core? I found [this Stack Overflow question](https://stackoverflow.com/questions/43813889/writing-to-a-session-variable-with-sessionstatebehavior-readonly).

VocaDbWeb.Core/Controllers/Api/ActivityEntryApiController.cs
```diff
@@ -0,0 +1,66 @@
...
+		/// <param name="fields">Optional fields.</param>
+		/// <param name="entryFields">Optional fields for entries.</param>
+		/// <param name="lang">Content language preference. Optional.</param>
+		/// <param name="sortRule">Sort rule. Optional.</param>
+		/// <returns>List of activity entries.</returns>
+		/// <remarks>
+		/// Entries are always returned sorted from newest to oldest.
+		/// Activity for deleted entries is not returned.
+		/// </remarks>
+		[HttpGet("")]
```
ycanardeau, line 51: Replaced `Route` with `HttpGet`, `HttpPost` and `HttpDelete` depending on the HTTP request method.

VocaDbWeb.Core/Startup.cs
```diff
@@ -50,7 +53,15 @@ public Startup(IConfiguration configuration)
...
 		public void ConfigureServices(IServiceCollection services)
 		{
-			services.AddControllersWithViews();
+			// Code from: https://stackoverflow.com/questions/51411693/null-response-returns-a-204/60858295#60858295
+			services.AddControllersWithViews(options =>
+			{
+				// remove formatter that turns nulls into 204 - No Content responses
+				// this formatter breaks Angular's Http response JSON parsing
+				options.OutputFormatters.RemoveType<HttpNoContentOutputFormatter>();
+			});
```
ycanardeau, line 62: `EntryTypesApiController.GetMappedTag` may return `null`.

VocaDbWeb.Core/Controllers/Api/ResourcesApiController.cs
riipah: If it's already using camelCase, then no need. I think that's the default now. Probably was some legacy code.

VocaDbWeb.Core/Controllers/HomeController.cs
riipah: I'm not even 100% are we using session state here? There might have been some session state regarding login/twitter/TempData.

According to the [documentation](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/app-state?view=aspnetcore-3.1), session state in ASP.NET Core is non-locking anyway, so no need to worry about the SessionStateBehavior.ReadOnly.

So to me it seems we don't need to worry about this attribute. There are some cases where TempData is used, session state might be necessary, but since it's non-locking, it should be ok by default.

VocaDbWeb.Core/Startup.cs
```diff
@@ -50,7 +57,29 @@ public Startup(IConfiguration configuration)
...
+					// this formatter breaks Angular's Http response JSON parsing
+					options.OutputFormatters.RemoveType<HttpNoContentOutputFormatter>();
+
+					options.Filters.Add<VoidAndTaskTo204NoContentFilter>();
+				})
+				.ConfigureApiBehaviorOptions(options =>
+				{
+					// Code from: https://docs.microsoft.com/en-us/aspnet/core/web-api/?view=aspnetcore-5.0#apicontroller-attribute
+					options.SuppressModelStateInvalidFilter = true;
+				})
```
ycanardeau, line 74: Disabled automatic 400 response.

VocaDbWeb.Core/Code/WebApi/VoidAndTaskTo204NoContentFilter.cs
```diff
@@ -0,0 +1,32 @@
...
+					context.HttpContext.Response.StatusCode = StatusCodes.Status204NoContent;
+			}
+		}
+
+		/// <inheritdoc/>
+		public void OnResultExecuted(ResultExecutedContext context)
+		{
+		}
+	}
+}
```
ycanardeau, line 32: For compatibility reasons.

VocaDbWeb.Core/Startup.cs
```diff
@@ -50,7 +57,29 @@ public Startup(IConfiguration configuration)
...
+				.AddNewtonsoftJson(options =>
+				{
+					options.SerializerSettings.ContractResolver = new CamelCasePropertyNamesContractResolver(); // All properties in camel case
+					options.SerializerSettings.Converters.Add(new StringEnumConverter());  // All enums as strings by default
+					options.SerializerSettings.NullValueHandling = NullValueHandling.Ignore;
+				});
```
ycanardeau, line 80: I decided to use `Newtonsoft.Json` instead of `System.Text.Json` for now.

VocaDbWeb.Core/Controllers/Api/ResourcesApiController.cs
```diff
@@ -0,0 +1,121 @@
...
+						ContractResolver = new DefaultContractResolver(),
+					};
+					// Code from: https://github.com/aspnet/Mvc/issues/8128#issuecomment-407214518
+					var options = context.HttpContext.RequestServices.GetRequiredService<IOptions<MvcOptions>>();
+					var mvcOptions = options.Value;
+					var jsonFormatter = new NewtonsoftJsonOutputFormatter(serializerSettings, ArrayPool<char>.Shared, mvcOptions);
+					objectResult.Formatters.Add(jsonFormatter);
+				}
+			}
+		}
```
ycanardeau, line 48: Implemented `DefaultCasingConfig`. It seems that `ResourcesApiController` uses PascalCase rather than camelCase.

The JSON response with `DefaultCasingConfig` (See https://vocadb.net/api/resources/en-US/?setNames%5B%5D=albumSortRuleNames):

```json
{
   "albumSortRuleNames":{
      "RatingAverage":"Rating average",
      "RatingTotal":"Total score",
      "ReleaseDate":"Release date",
      "CollectionCount":"Collection count",
      "AdditionDate":"Addition date",
      "Name":"Name"
   }
}
```

The JSON response without `DefaultCasingConfig`:

```json
{
   "albumSortRuleNames":{
      "ratingAverage":"Rating average",
      "ratingTotal":"Total score",
      "releaseDate":"Release date",
      "collectionCount":"Collection count",
      "additionDate":"Addition date",
      "name":"Name"
   }
}
```

VocaDbWeb.Core/Code/WebApi/RestrictBannedIPAttribute.cs
```diff
@@ -0,0 +1,31 @@
...
+
+			if (!ipRules.IsAllowed(host))
+			{
+				// Code from: https://stackoverflow.com/questions/31617345/what-is-the-asp-net-core-mvc-equivalent-to-request-requesturi/40721652#40721652
+				s_log.Warn($"Restricting banned host '{host}' for '{new Uri(context.HttpContext.Request.GetDisplayUrl())}'.");
+
+				context.Result = new ForbidResult();
+			}
+		}
+	}
```
ycanardeau, line 30: Implemented `RestrictBannedIPAttribute`, but need to be tested later on.

VocaDbWeb.Core/Controllers/Api/IPRuleApiController.cs
```diff
@@ -0,0 +1,79 @@
...
+			});
+		}
+
+		[HttpPost("")]
+		public ActionResult<bool> PostNewIPRule(IPRule rule)
+		{
+			_userContext.VerifyPermission(PermissionToken.ManageIPRules);
+
+			if (string.IsNullOrEmpty(rule?.Address))
+				return BadRequest();
```
ycanardeau, line 66: Replaced `HttpBadRequestException` with `BadRequest`.

VocaDbWeb.Core/Controllers/Api/PVApiController.cs
```diff
@@ -0,0 +1,70 @@
...
+		{
+			if (string.IsNullOrEmpty(pvUrl))
+				return BadRequest();
+
+			var result = await _pvParser.ParseByUrlAsync(pvUrl, getTitle, _permissionContext);
+
+			if (!result.IsOk)
+			{
+				var msg = result.Exception.Message;
+				return BadRequest(msg);
```
ycanardeau, line 63: The original code was:

```csharp
throw new HttpResponseException(new HttpResponseMessage(HttpStatusCode.BadRequest)
{
    ReasonPhrase = msg,
    Content = new StringContent(msg)
});
```

VocaDbWeb.Core/Startup.cs
riipah: You should mention in code comments why is this here.

VocaDbWeb.Core/Controllers/Api/ResourcesApiController.cs
riipah: I guess there was a reason ResourceApi was left PascalCase. But yeah, default should be camelCase.

ycanardeau: Added `[Obsolete]` to `DefaultCasingConfig`.

VocaDbWeb.Core/Startup.cs
```diff
@@ -179,6 +212,15 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
...
 			});
+
+			app.UseCacheOutput();
+
+			// Code from: https://docs.microsoft.com/en-us/aspnet/core/tutorials/getting-started-with-swashbuckle?view=aspnetcore-5.0&tabs=visual-studio
+			app.UseSwagger();
+			app.UseSwaggerUI(c =>
+			{
+				c.SwaggerEndpoint("/swagger/v1/swagger.json", "VocaDB Web API V1");
+			});
```
ycanardeau, line 223: We don't use custom assets for now.

VocaDbWeb.Core/Code/WebApi/RequireSslAttribute.cs
```diff
@@ -0,0 +1,28 @@
...
+			return;
+#endif
+
+			if (!IsSSL(context.HttpContext.Request))
+			{
+				// 403.4 - SSL required.
+				context.Result = new ForbidResult("This API requires SSL");
+			}
+		}
+	}
```
ycanardeau, line 27: Needs to be tested.

VocaDbWeb.Core/Code/WebApi/AuthenticatedCorsApiAttribute.cs
```diff
@@ -0,0 +1,57 @@
...
+			var origins = AppConfig.AllowedCorsOrigins;
+			if (!string.IsNullOrEmpty(origins))
+			{
+				if (origins != "*")
+					Array.ForEach(origins.Split(','), _policy.Origins.Add);
+			}
+		}
+
+		public Task<CorsPolicy> GetPolicyAsync(HttpContext context, string policyName) => Task.FromResult(_policy);
+	}
```
ycanardeau, line 56: Needs to be tested.

VocaDbWeb.Core/Controllers/Api/SampleObjects.cs
```diff
@@ -0,0 +1,178 @@
...
+using VocaDb.Model.Domain.Artists;
+using VocaDb.Model.Domain.ExtLinks;
+using VocaDb.Model.Domain.Globalization;
+using VocaDb.Model.Domain.PVs;
+using VocaDb.Model.Domain.Songs;
+using VocaDb.Model.Service;
+
+namespace VocaDb.Web.Areas.HelpPage.App_Start
+{
+	public static class SampleObjects
```
ycanardeau, line 22: What is the purpose of this class?

VocaDbWeb.Core/Helpers/ViewRenderService.cs
```diff
@@ -0,0 +1,70 @@
...
+				viewDictionary,
+				new TempDataDictionary(actionContext.HttpContext, _tempDataProvider),
+				sw,
+				new HtmlHelperOptions()
+			);
+
+			await viewResult.View.RenderAsync(viewContext);
+			return sw.ToString();
+		}
+	}
```
ycanardeau, line 69: Replacement for `RazorHelper`. Needs to be tested.

VocaDbWeb.Core/Startup.cs
```diff
@@ -137,6 +169,7 @@ public void ConfigureContainer(ContainerBuilder builder)
 			builder.RegisterType<LaravelMixHelper>().AsSelf();
 			builder.RegisterType<Login>().AsSelf();
 			builder.RegisterType<PVHelper>().AsSelf();
+			builder.RegisterType<ViewRenderService>().As<IViewRenderService>();
```
ycanardeau, line 172: When should I use `SingleInstance`?

VocaDbWeb.Core/Code/WebApi/RequireSslAttribute.cs
riipah: In ASP.NET Core you can make the whole app require SSL, which would be okay and probably simpler, now that it's all using SSL anyway. This attribute is from the time when only certain pages were SSL. Even in development you could use SSL, because IIS Express supports it using self-signed certificate.

VocaDbWeb.Core/Code/WebApi/AuthenticatedCorsApiAttribute.cs
riipah: ASP.NET Core [includes its own CORS settings](https://docs.microsoft.com/en-us/aspnet/core/security/cors?view=aspnetcore-5.0). No need for this custom stuff anymore.

riipah: This should probably be refactored to use the standard builder/middleware syntax later.

VocaDbWeb.Core/Controllers/Api/SampleObjects.cs
riipah: In the API Explorer that we were using before Swagger, there was possibility to specify custom sample data for API requests. Don't know if Swagger supports that. Actually I remembered that we were already using this for Swagger, but guess not.

VocaDbWeb.Core/Startup.cs
riipah: When there should be only one instance of a class per application. If the class has no state (or you want shared state) and doesn't depend on user or culture (language), then it can usually be singleton. Singleton has better performance because of less allocations.

VocaDbWeb.Core/Controllers/Api/ResourcesApiController.cs
```diff
@@ -0,0 +1,123 @@
...
+
+		private const int CacheDuration = Model.Domain.Constants.SecondsInADay;
+
+		private readonly Dictionary<string, ResourceManager> allSets = new()
```
riipah, line 54: Would be good to move this away from the controller. Maybe to the resources project?

VocaDbWeb.Core/Controllers/Api/ResourcesApiController.cs
```diff
@@ -0,0 +1,123 @@
...
+		/// Gets a number of resource sets for a specific culture.
+		/// </summary>
+		/// <param name="cultureCode">Culture code, for example "en-US" or "fi-FI".</param>
+		/// <param name="setNames">Names of resource sets to be returned. More than one value can be specified. For example "artistTypeNames"</param>
+		/// <returns>Resource sets. Cannot be null.</returns>
+		[HttpGet("{cultureCode}")]
+		[CacheOutput(ClientTimeSpan = CacheDuration, ServerTimeSpan = CacheDuration)]
+		public Dictionary<string, Dictionary<string, string>> GetList(string cultureCode, [FromQuery(Name = "setNames[]")] string[] setNames)
+		{
+			if (setNames == null || !setNames.Any())
```
riipah, line 101: This code too, the less code there is in controllers, the better.

VocaDbWeb.Core/Controllers/Api/ResourcesApiController.cs
ycanardeau: Some resources are located in `VocaDb.Model` not in `VocaDb.Resources`. We need some refactoring before doing this. Let's keep it as is for now.

VocaDbWeb.Core/Code/ErrorLogger.cs
```diff
@@ -0,0 +1,61 @@
...
+		public static string RequestInfo(string msg, HttpRequest request)
+		{
+			var userHostAddress = request.HttpContext.Connection.RemoteIpAddress;
+			var userHostName = request.GetTypedHeaders().Host;
+			var httpMethod = request.Method;
+			var pathAndQuery = request.GetEncodedPathAndQuery();
+			var userAgent = request.Headers[HeaderNames.UserAgent];
+			var urlReferrer = request.GetTypedHeaders().Referer;
+			return $"{msg} '{userHostAddress}' [{userHostName}], URL {httpMethod} '{pathAndQuery}', UA '{userAgent}', referrer '{urlReferrer}'";
+		}
```
ycanardeau, line 59: Implemented `ErrorLogger.RequestInfo`.

VocaDbWeb.Core/Controllers/UserController.cs
```diff
@@ -0,0 +1,150 @@
...
+					var claims = new List<Claim>
+					{
+						new Claim(ClaimTypes.Name, user.Name),
+					};
+					var claimsIdentity = new ClaimsIdentity(claims, CookieAuthenticationDefaults.AuthenticationScheme);
+					var authProperties = new AuthenticationProperties
+					{
+						IsPersistent = model.KeepLoggedIn,
+					};
+					await HttpContext.SignInAsync(CookieAuthenticationDefaults.AuthenticationScheme, new ClaimsPrincipal(claimsIdentity), authProperties);
```
ycanardeau, line 105: Sign in.

VocaDbWeb.Core/Controllers/UserController.cs
```diff
@@ -0,0 +1,150 @@
...
+				SaveErrorsToTempData();
+				return Redirect(VocaUriBuilder.Absolute(Url.Action("Login", new { model.ReturnUrl })));
+			}
+
+			return View(model);
+		}
+
+		public async Task<ActionResult> Logout()
+		{
+			await HttpContext.SignOutAsync(CookieAuthenticationDefaults.AuthenticationScheme);
```
ycanardeau, line 146: Sign out.

VocaDbWeb.Core/Helpers/WebHelper.cs
```diff
@@ -1,12 +1,37 @@
...
+		{
+			if (request.GetTypedHeaders().AcceptLanguage == null || !request.GetTypedHeaders().AcceptLanguage.Any())
+				return string.Empty;
+
+			return GetUserLanguageCodes(request)
+				.Select(l => l.GetCultureInfoSafe())
+				.Where(l => l != null && InterfaceLanguage.IsValidUserInterfaceCulture(l))
+				.Select(l => l.Name)
+				.FirstOrDefault() ?? string.Empty;
+		}
```
ycanardeau, line 31: Needs to be tested.

VocaDbWeb.Core/Middleware/VocaDbPrincipalMiddleware.cs
```diff
@@ -0,0 +1,61 @@
...
+		{
+			_next = next;
+		}
+
+		private static bool IsAjaxRequest(HttpRequest request)
+		{
+			ParamIs.NotNull(() => request);
+
+			return request.Headers["X-Requested-With"] == "XMLHttpRequest";
+		}
```
ycanardeau, line 27: Needs to be tested.

VocaDbWeb.Core/Middleware/VocaDbPrincipalMiddleware.cs
```diff
@@ -0,0 +1,61 @@
...
+				loginManager.InitLanguage();
+			}
+			catch (Exception x)
+			{
+				// TODO: this should be processed using normal exception handling.
+				ErrorLogger.LogException(context.Request, x, LogLevel.Fatal);
+			}
+
+			return _next(context);
+		}
```
ycanardeau, line 54: Replacement for `Application_AuthenticateRequest`.

VocaDbWeb.Core/Middleware/VocaDbPrincipalMiddleware.cs
```diff
@@ -0,0 +1,61 @@
...
+using Microsoft.AspNetCore.Http;
+using NLog;
+using VocaDb.Model;
+using VocaDb.Model.Service;
+using VocaDb.Model.Service.Security;
+using VocaDb.Web.Code;
+
+namespace VocaDb.Web.Middleware
+{
+	public class VocaDbPrincipalMiddleware
```
ycanardeau, line 13: Is there a better name for this class?

VocaDbWeb.Core/Controllers/ControllerBase.cs
```diff
@@ -0,0 +1,54 @@
...
+
+		protected void RestoreErrorsFromTempData()
+		{
+			// TODO: implement
+		}
+
+		protected void SaveErrorsToTempData()
+		{
+			// TODO: implement
+		}
```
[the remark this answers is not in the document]
riipah: Yeah it seems these methods are only needed if ReturnToMainSite is true. That ReturnToMainSite was there because the login page used HTTPS while the rest of the site was HTTP (because NicoNico took a looong time to add HTTPS support). Anyway as you said, these are **probably** not needed anymore, but I'm not 100% sure.

VocaDbWeb.Core/Controllers/UserController.cs
```diff
@@ -0,0 +1,149 @@
...
+					{
+						IsPersistent = model.KeepLoggedIn,
+					};
+					await HttpContext.SignInAsync(CookieAuthenticationDefaults.AuthenticationScheme, new ClaimsPrincipal(claimsIdentity), authProperties);
+
+					string redirectUrl = null;
+					// TODO: implement
+					/*try
+					{
+						redirectUrl = FormsAuthentication.GetRedirectUrl(model.UserName, true);
```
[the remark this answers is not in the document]
riipah: If you try to access a page that requires login, such as Admin page, you get redirected to login page and the admin page URL will be in GetRedirectUrl, so after login you'd be redirected to the Admin page, where you wanted to get in the first place. This is not terribly important on VocaDB, because on normal pages the login happens in a popup dialog so redirecting is not needed. It's probably ok to leave it out initially, but should be reimplemented eventually.

VocaDbWeb.Core/Middleware/VocaDbPrincipalMiddleware.cs
riipah: In any case this class should be temporary and eventually be replaced with something that fits ASP.NET Core authentication better.

VocaDbWeb.Core/Controllers/UserController.cs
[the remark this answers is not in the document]
ycanardeau: It seems that it's already done by the `ReturnUrl` query string?

VocaDbWeb.Core/Code/JsonModelBinder.cs
```diff
@@ -0,0 +1,48 @@
...
+			catch (JsonReaderException x)
+			{
+				s_log.Error(x, "Unable to process JSON, content is " + value);
+				throw;
+			}
+
+			bindingContext.Result = ModelBindingResult.Success(obj);
+
+			return Task.CompletedTask;
+		}
```
ycanardeau, line 46: Needs to be tested.

VocaDbWeb.Core/Controllers/AdminController.cs
```diff
@@ -0,0 +1,326 @@
...
+		{
+			PermissionContext.VerifyPermission(PermissionToken.ManageIPRules);
+
+			var rules = _otherService.GetIPRules();
+			return View(rules);
+		}
+
+		[Authorize]
+		[HttpPost]
+		public ActionResult ManageIPRules([ModelBinder(BinderType = typeof(JsonModelBinder))] IPRule[] rules)
```
ycanardeau, line 176: Replaced `[FromJson]` with `[ModelBinder(BinderType = typeof(JsonModelBinder))]`.

VocaDbWeb.Core/Models/Tag/TagEditViewModel.cs
```diff
@@ -0,0 +1,150 @@
...
+using VocaDb.Model.Domain.Globalization;
+using VocaDb.Model.Domain.Security;
+using VocaDb.Model.Domain.Tags;
+using VocaDb.Web.Code;
+using VocaDb.Web.Code.Exceptions;
+using VocaDb.Web.Helpers;
+
+namespace VocaDb.Web.Models.Tag
+{
+	// TODO: implement [PropertyModelBinder]
```
ycanardeau, line 21: We no longer need this anymore?

VocaDbWeb.Core/Models/Tag/TagEditViewModel.cs
```diff
@@ -0,0 +1,150 @@
...
+		public string CategoryName { get; set; }
+
+		public string CurrentName { get; set; }
+
+		public bool Deleted { get; set; }
+
+		public ContentLanguageSelection DefaultNameLanguage { get; set; }
+
+		[Display(Name = "Description")]
+		[ModelBinder(BinderType = typeof(JsonModelBinder))]
```
ycanardeau, line 61: Replaced `FromJson` with `[ModelBinder(BinderType = typeof(JsonModelBinder))]`.

VocaDbWeb.Core/Controllers/TagController.cs
```diff
@@ -0,0 +1,277 @@
...
+				ModelState.AddModelError("targetTagId", "Tag must be selected");
+				return Merge(id);
+			}
+
+			_queries.Merge(id, targetTagId.Value);
+
+			return RedirectToAction("Edit", new { id = targetTagId.Value });
+		}
+
+		[ResponseCache(Location = ResponseCacheLocation.Any, Duration = 3600)]
```
ycanardeau, line 240: Replaced `OutputCache` with `ResponseCache`.

VocaDbWeb.Core/Controllers/TagController.cs
```diff
@@ -0,0 +1,277 @@
...
+		[HttpPost]
+		[Authorize]
+		public ActionResult Edit(TagEditViewModel model)
+		{
+			var coverPicUpload = Request.Form.Files["thumbPicUpload"];
+			UploadedFileContract uploadedPicture = null;
+			if (coverPicUpload != null && coverPicUpload.Length > 0)
+			{
+				CheckUploadedPicture(coverPicUpload, "thumbPicUpload");
+				uploadedPicture = new UploadedFileContract { Mime = coverPicUpload.ContentType, Stream = coverPicUpload.OpenReadStream() };
```
ycanardeau, line 170: Replaced `InputStream` with `OpenReadStream()`.

VocaDbWeb.Core/Controllers/TagController.cs
```diff
@@ -0,0 +1,277 @@
...
+			var contract = _queries.GetTagForEdit(model.Id);
+			model.CopyNonEditableProperties(contract, PermissionContext);
+			return View("Edit", model);
+		}
+
+		[HttpPost]
+		[Authorize]
+		public ActionResult Edit(TagEditViewModel model)
+		{
+			var coverPicUpload = Request.Form.Files["thumbPicUpload"];
```
ycanardeau, line 165: Replaced `Request.Files["thumbPicUpload"]` with `Request.Form.Files["thumbPicUpload"]`.

VocaDbWeb.Core/Controllers/TagController.cs
```diff
@@ -0,0 +1,277 @@
...
+			UploadedFileContract uploadedPicture = null;
+			if (coverPicUpload != null && coverPicUpload.Length > 0)
```
ycanardeau, line 167: Replaced `ContentLength` with `Length`.

VocaDbWeb.Core/Controllers/UserController.cs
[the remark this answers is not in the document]
riipah: I think ReturnUrl is added by VocaDB to override the ASP.NET redirectUrl? But most likely one of them is not needed.

VocaDbWeb.Core/Models/Tag/TagEditViewModel.cs
riipah: You mean PropertyModelBinderAttribute? I think the purpose was to activate FromJsonAttributes. But yeah, might be that in ASP.NET Core it's not needed anymore.

VocaDbWeb.Core/Controllers/ControllerBase.cs
```diff
@@ -0,0 +1,150 @@
...
+					return null;
+				}
+
+				if (!ImageHelper.IsValidImageExtension(pictureUpload.FileName))
+				{
+					ModelState.AddModelError(fieldName, "Picture format is not valid.");
+					return null;
+				}
+
+				pictureData = new EntryPictureFileContract(pictureUpload.OpenReadStream(), pictureUpload.ContentType, (int)pictureUpload.Length, purpose);
```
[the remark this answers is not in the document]
riipah: Should be, we don't have that huge pictures.

VocaDbWeb.Core/Controllers/ControllerBase.cs
```diff
@@ -0,0 +1,150 @@
...
+				pictureData.OriginalFileName = pictureUpload.FileName;
+			}
+
+			return pictureData;
+		}
+
+		protected ActionResult HttpStatusCodeResult(HttpStatusCode code, string message)
+		{
+			Response.StatusCode = (int)code;
+			// TODO: implement Response.StatusDescription = message;
```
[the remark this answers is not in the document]
riipah: The best solution would be to not customize StatusDescription at all. The response details should all be in content/body, or possibly in custom response header.

VocaDbWeb.Core/Controllers/ExtController.cs
```diff
@@ -0,0 +1,193 @@
...
+using VocaDb.Model.Domain.Images;
+using VocaDb.Model.Helpers;
+using VocaDb.Model.Service;
+using VocaDb.Model.Utils;
+using VocaDb.Web.Helpers;
+using VocaDb.Web.Models.Ext;
+
+namespace VocaDb.Web.Controllers
+{
+	public class ExtController : ControllerBase
```
ycanardeau, line 24: Removed `SessionStateBehavior.Disabled`.

VocaDbWeb.Core/Controllers/ControllerBase.cs
```diff
@@ -0,0 +1,254 @@
...
+			{
+				//var encoded = Url.Encode(title);
+				// Note: there is no good way to encode content-disposition filename (see http://stackoverflow.com/a/216777)
+				Response.Headers.Add("content-disposition", $"inline;filename=\"{title}{ext}\"");
+			}
+
+			return File(pictureData.Bytes, pictureData.Mime);
+		}
+
+		protected ActionResult LowercaseJson(object obj) => Content(JsonHelpers.Serialize(obj), "application/json");
```
ycanardeau, line 211: Can we implement `LowercaseJson` like this? [The original implementation](https://github.com/VocaDB/vocadb/blob/2b78a2b42702e20dbc870ea58ad8c832c1cfcea9/VocaDbWeb/Controllers/ControllerBase.cs#L214) makes use of `JsonNetResult`.

VocaDbWeb.Core/Models/Album/AlbumEditViewModel.cs
```diff
@@ -0,0 +1,90 @@
...
+		public EntryStatus[] AllowedEntryStatuses { get; set; }
+
+		public bool CanDelete { get; set; }
+
+		public string DiscTypeDescriptions { get; set; }
+
+		public bool Draft => Album != null && Album.Status == EntryStatus.Draft;
+
+		[ModelBinder(BinderType = typeof(JsonModelBinder))]
+		// TODO: implement [AllowHtml]
```
ycanardeau, line 57: Why is the `AllowHtml` attribute needed here?

VocaDbWeb.Core/Controllers/ControllerBase.cs
```diff
@@ -0,0 +1,254 @@
...
+			};
+
+			Response.Headers[HeaderNames.ETag] = $"{contract.EntryType}{contract.EntryId}v{contract.Version}";
+
+			// Cached version indicated by the "v" request parameter.
+			// If no version is specified, assume no caching.
+			if (contract.Version > 0 && !string.IsNullOrEmpty(Request.Query["v"]))
+				cacheControl.MaxAge = PictureCacheDuration;
+
+			Response.GetTypedHeaders().CacheControl = cacheControl;
```
ycanardeau, line 84: I'm not 100% sure if this is the right way to implement this method.

VocaDbWeb.Core/Code/Web/AspNetCoreHttpPostedFile.cs
```diff
@@ -0,0 +1,22 @@
...
+			_file = file;
+		}
+
+		private readonly IFormFile _file;
+
+		public string ContentType => _file.ContentType;
+
+		public string FileName => _file.FileName;
+
+		public void SaveAs(string path) => _file.CopyTo(new FileStream(path, FileMode.Create));
```
ycanardeau, line 20: Needs to be tested.

VocaDbWeb.Core/Controllers/SongController.cs
```diff
@@ -0,0 +1,590 @@
...
+				song => Url.Action("Details", new { id = song.Id }));
+
+			return new FeedResult(new Atom10FeedFormatter(feed));
+		}*/
+
+		// TODO: implement
+		/*public FeedResult LatestVideos()
+		{
+			return Feed(new IndexRouteParams { onlyWithPVs = true, pageSize = 20, sort = SongSortRule.AdditionDate });
+		}*/
```
ycanardeau, line 368: Needs to be implemented. Perhaps related to #713.

VocaDbWeb.Core/Helpers/ViewHelper.cs
```diff
@@ -207,7 +208,38 @@ public static RouteValueDictionary GetRouteValueDictionary(SongContract contract
...
+				tdTag.InnerHtml.AppendHtml(contentFunc(item));
+				trTag.InnerHtml.AppendHtml(tdTag);
+				i++;
+			}
+
+			if (trTag != null)
+				tableTag.InnerHtml.AppendHtml(trTag);
+
+			return tableTag;
+		}
```
ycanardeau, line 242: Replaced `InnerHtml +=` with `InnerHtml.AppendHtml`. Returns `TagBuilder` directly without converting to string.

VocaDbWeb.Core/Startup.cs
```diff
@@ -171,14 +213,58 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
...
 
+			app.UseVocaDbPrincipal();
+
 			app.UseEndpoints(endpoints =>
 			{
+				const string Numeric = "[0-9]+";
+
+				// Ignored files
+				// TODO: implement routes.IgnoreRoute("{resource}.axd/{*pathInfo}");
+				// TODO: implement routes.IgnoreRoute("favicon.ico");
```
ycanardeau, line 227: Do we need to ignore these two routes?

VocaDbWeb.Core/Code/IdNotNumberConstraint.cs
```diff
@@ -0,0 +1,14 @@
...
+namespace VocaDb.Web.Code
+{
+	public class IdNotNumberConstraint : IRouteConstraint
+	{
+		public bool Match(HttpContext? httpContext, IRouter? route, string routeKey, RouteValueDictionary values, RouteDirection routeDirection)
+		{
+			var val = values[routeKey]?.ToString();
+			return !int.TryParse(val, out _);
+		}
+	}
```
ycanardeau, line 13: Needs to be tested.

VocaDbWeb.Core/Startup.cs
```diff
@@ -50,7 +59,43 @@ public Startup(IConfiguration configuration)
...
+					// this formatter breaks Angular's Http response JSON parsing
+					options.OutputFormatters.RemoveType<HttpNoContentOutputFormatter>();
+
+					options.Filters.Add<VoidAndTaskTo204NoContentFilter>();
+				})
+				.ConfigureApiBehaviorOptions(options =>
+				{
+					// Code from: https://docs.microsoft.com/en-us/aspnet/core/web-api/?view=aspnetcore-5.0#disable-inference-rules
+					// Disable inference rules. This prevents API controllers from returning an UnsupportedMediaTypeResult (415).
+					options.SuppressInferBindingSourcesForParameters = true;
```
ycanardeau, line 76: Is it okay to disable inference rules?

riipah: Could you comment here why is this needed?

VocaDbWeb.Core/Startup.cs
riipah: If it's not causing errors, no

VocaDbWeb.Core/Controllers/SongController.cs
riipah: These are not related to #713. These are RSS feeds that are being used by some people to follow new songs added.

VocaDbWeb.Core/Controllers/ControllerBase.cs
riipah: Looks okay...

VocaDbWeb.Core/Models/Album/AlbumEditViewModel.cs
riipah: Album description at least may contain characters that are interpreted as HTML. Otherwise ASP.NET request validation would block all ```<``` and ```>``` at least.

VocaDbWeb.Core/Controllers/ControllerBase.cs
riipah: If all JSON is serialized as camelCase by default, then yes.

VocaDbWeb.Core/Views/Admin/ViewAuditLog.cshtml
```diff
@@ -0,0 +1,111 @@
...
+				$("#start").val(parseInt($("#start").val()) + 200);
+			});
+			return false;
+		});
+
+		@* HACK: replacement for Html.RenderAction. *@
+		@* Code from: https://stackoverflow.com/questions/41353874/equivalent-of-html-renderaction-in-asp-net-core/45494932#45494932 *@
+		$.get('@Url.Action("AuditLogEntries", Model)', null, function (content) {
+			$("#logEntries").html(content);
+		});
```
ycanardeau, line 107: Replacement for `RenderAction`.

VocaDbWeb.Core/Controllers/UserController.cs
```diff
@@ -0,0 +1,918 @@
...
+			_repository = repository;
+			_otherService = otherService;
+			_messageQueries = messageQueries;
+			_ipRuleManager = ipRuleManager;
+			_config = config;
+			_markdownParser = markdownParser;
+			_loginManager = loginManager;
+		}
+
+		[HttpPost]
```
ycanardeau, line 98: Replaced `[AcceptVerbs(HttpVerbs.Post)]` with `[HttpPost]`.

VocaDbWeb.Core/Controllers/StatsController.cs
```diff
@@ -0,0 +1,805 @@
...
+
+		private ActionResult DateLineChartWithAverage(string title, string pointsTitle, string yAxisTitle, ICollection<Tuple<DateTime, int>> points,
+			bool average = true)
+		{
+			Response.GetTypedHeaders().CacheControl = new CacheControlHeaderValue
+			{
+				Public = true,
+				MaxAge = TimeSpan.FromDays(1),
+			};
+			// TODO: implement Response.Cache.SetSlidingExpiration(true);
```
ycanardeau, line 127: How can I perform sliding expiration?

VocaDbWeb.Core/Controllers/StatsController.cs
```diff
@@ -0,0 +1,805 @@
...
+		public ActionResult CumulativeSongsPublished(DateTime? cutoff)
+		{
+			var values = _songAggregateQueries.SongsOverTime(TimeUnit.Month, false, cutoff, a => a.PVs.PVs.Any(), a => a.PVs.PVs.Count == 0).First();
+
+			var points = values.CumulativeSum();
+
+			return DateLineChartWithAverage("Cumulative songs published per day", "Songs", "Number of songs", points, false);
+		}
+
+		[ResponseCache(Duration = ClientCacheDurationSec, VaryByQueryKeys = new[] { "cutoff" })]
```
ycanardeau, line 440: Replaced `VaryByParam` with `VaryByQueryKeys`.

VocaDbWeb.Core/Controllers/StatsController.cs
```diff
@@ -0,0 +1,805 @@
...
+		}*/
+		private T GetCachedReport<T>() where T : class => throw new NotImplementedException();
+
+		// TODO: implement
+		/*private void SaveCachedReport<T>(T data) where T : class
+		{
+			var name = ControllerContext.RouteData.Values["action"];
+			_context.Cache.Add("report_" + name, data, null, Cache.NoAbsoluteExpiration, TimeSpan.FromDays(1), CacheItemPriority.Default, null);
+		}*/
+		private void SaveCachedReport<T>(T data) where T : class => throw new NotImplementedException();
```
ycanardeau, line 57: How can I implement these two methods? Should we use `ObjectCache`?

VocaDbWeb.Core/Controllers/AlbumController.cs
```diff
@@ -290,20 +290,13 @@ public async Task<ActionResult> Edit(AlbumEditViewModel viewModel)
 			var coverPicUpload = Request.Form.Files["coverPicUpload"];
 			var pictureData = ParsePicture(coverPicUpload, "CoverPicture", ImagePurpose.Main);
 
-			if (coverPicUpload == null)
-			{
-				AddFormSubmissionError("Cover picture was null");
-			}
```
ycanardeau: `coverPicUpload` may be `null` here.

VocaDbWeb.Core/Controllers/StatsController.cs
riipah: Maybe just ResponseCache? But yes, the old ASP.NET cache can always be replaced with ObjectCache.

VocaDbWeb.Core/Controllers/StatsController.cs
riipah: It's not that important. Max age is enough here.

VocaDbWeb.Core/Views/Admin/ViewAuditLog.cshtml
riipah: Maybe you can add comment to replace with view components in the future?

VocaDbWeb.Core/Startup.cs
riipah: But **why** is it returning unsupported media type?

ycanardeau: I don't know why. I found some possibly related issues:

- aspnet/Mvc#8116
- dotnet/aspnetcore#9932

Currently, [AlbumApiController.PostReview](https://github.com/VocaDB/vocadb/blob/817001d0fb8ebe121b0e4ab53dc8211941f5a3da/VocaDbWeb.Core/Controllers/Api/AlbumApiController.cs#L240) throws an `ObjectNotFoundException` since the album ID is passed not only to `id`, but also to `reviewContract.id`. We can solve this by adding the `[FromForm]` attribute to `AlbumReviewContract reviewContract`. Should we use this attribute instead of disabling inference rules? Originally, I tried to solve `Unsupported Media Type` with this attribute, I was not sure if this was the right way.

ycanardeau: I don't know why. I found some possibly related issues:

- aspnet/Mvc#8116
- dotnet/aspnetcore#9932

Currently, [AlbumApiController.PostReview](https://github.com/VocaDB/vocadb/blob/817001d0fb8ebe121b0e4ab53dc8211941f5a3da/VocaDbWeb.Core/Controllers/Api/AlbumApiController.cs#L240) throws an `ObjectNotFoundException` since the album ID is passed not only to `id`, but also to `reviewContract.id`. We can solve this by adding the `[FromForm]` attribute to `AlbumReviewContract reviewContract`. Should we use this attribute instead of disabling inference rules? Originally, I tried to solve `Unsupported Media Type` with this attribute, but I was not sure if this was the right way.

riipah: Is the problem that it's binding the ID to wrong field? Adding the attribute sounds like more proper way of fixing it. Setting this global SuppressInferBindingSourcesForParameters sounds like a last resort workaround.

VocaDbWeb.Core/Startup.cs
riipah: By the way, why is it sending CategoryName?

VocaDbWeb.Core/Startup.cs
ycanardeau: Yes, I also tried `[FromBody]`, but in vain. Maybe related to [this SO question](https://stackoverflow.com/questions/49228906/415-unsupported-media-type-asp-net-core). Actually, I need to replace `[FromBody]` with `[FromForm]` in `UserApiController`.

VocaDbWeb.Core/Startup.cs
ycanardeau: Because [TagAutoComplete.ts](https://github.com/VocaDB/vocadb/blob/e96b44e6c7970e5300e6780aad2d39cf91a65509/VocaDbWeb/Scripts/KnockoutExtensions/TagAutoComplete.ts#L26) is sending it. We should remove it at some point though.

riipah: Sounds like it would be very easy to fix TagAutoComplete not to send it.

VocaDbWeb.Core/Startup.cs
ycanardeau: How about overriding `$.post` in `libs.js`?

```javascript
$.post = function (url, data, success, dataType) {
	return $.ajax({ url: url, type: "POST", dataType: dataType, contentType: "application/json", data: ko.toJSON(data), success: success });
};
```

ycanardeau: How about overriding `$.post` in `libs.js`?

```javascript
$.post = function (url, data, success, dataType) {
	return $.ajax({ url: url, type: "POST", dataType: dataType, contentType: "application/json", data: ko.toJSON(data), success: success });
};
```

VocaDbWeb.Core/wwwroot/Scripts/Repositories/AdminRepository.ts
```diff
@@ -6,7 +6,7 @@ import UrlMapper from '../Shared/UrlMapper';
         constructor(private urlMapper: UrlMapper) { }
 
 		public addIpToBanList = (rule: IPRuleContract, callback: (result: boolean) => void) => {
-			return $.post(this.urlMapper.mapRelative("/api/ip-rules"), rule, callback);
+			return $.postJSON(this.urlMapper.mapRelative("/api/ip-rules"), rule, callback);
```
ycanardeau, line 9: Replaced `$.post` with `$.postJSON`.

VocaDbWeb.Core/wwwroot/Scripts/libs.js
```diff
@@ -28,3 +28,7 @@ window._ = require('lodash');
 require('qtip2');
 
 window.marked = require('marked');
+
+$.postJSON = function (url, data, success, dataType) {
+	return $.ajax({ url: url, type: "POST", dataType: dataType, contentType: "application/json", data: ko.toJSON(data), success: success });
+};
```
ycanardeau, line 34: Implemented `$.postJSON`.

VocaDbWeb.Core/wwwroot/Scripts/typings/jquery.d.ts
```diff
@@ -0,0 +1,8 @@
+
+interface JQueryStatic {
+
+	postJSON(url: string, success?: (data: any, textStatus: string, jqXHR: JQueryXHR) => any, dataType?: string): JQueryXHR;
+
+	postJSON(url: string, data?: Object|string, success?: (data: any, textStatus: string, jqXHR: JQueryXHR) => any, dataType?: string): JQueryXHR;
+
+}
```
ycanardeau, line 8: This is the declaration file.

VocaDbWeb.Core/Startup.cs
ycanardeau: I've removed `CategoryName`. Should we re-enable automatic 400 response?

VocaDbWeb.Core/Startup.cs
ycanardeau: Thanks for the explanation!

VocaDbWeb.Core/Startup.cs
riipah: I think it would be best to re-enable it for now. Going with default config is always best if possible. And it may help us notice other issues, that's the intention anyway.

riipah:
By the way, once it's mostly working, I think it would be ok to merge this PR, even if some controllers and minor features are missing. Maybe even merge it to future branch. You don't need to do everything in 1 PR. It will be easier to review in smaller parts anyway. I only had the web-core branch because the frontend changes in future were not ready to be merged to master, but now that they are merged, we can use future for the .NET Core conversion, even if incomplete.

ycanardeau:
Got it, thank you for your review!

riipah:
No problem, thanks for coding it!

Issue #458, opened 2018-12-30
Add artist type for Synthesizer V


riipah:
Currently following the situation, seeing if there's going to be more voicebanks and songs.

andreoda:
(At least 5 root voicebanks):
Voidol - 18 root voicebanks
NIAONiao - 17 root voicebanks
Voiceroid - 13 root voicebanks
chipspeech - 12 root voicebanks
Synthesizer V - 9 root voicebanks
Muta - 8 root voicebanks
Alter/Ego - 6 root voicebanks
Sharpkey - 6 root voicebanks
Microsoft Speech API - 5 root voicebanks
Sinsy - 5 root voicebanks

There should be at least enough root voicebanks for an artist type field to make sense.
If we consider Synth V then we should consider the ones above as well.

riipah:
I would also consider the number of songs/albums, not just number of voicebanks. And some of those are discontinued as far as I know. I think chipspeech and Alter/Ego said they are not making any more voices.

andreoda:
How about making this available for editors to add. For example just like adding a new base/main/parent tag like the ones we have (Genres, Themes, etc.). 
We could simply add new artist "parent" types manually and the system automatically changes the colors of the aquare field. 
For example Cevio has a green color. The system would randomly pick a color that doesn't exists as an artist type already. If we decide to delete an artist type, then the color will be available again...or something like that. I don't know how complex that would be though.

ycanardeau:
@andreoda 

>There should be at least enough root voicebanks for an artist type field to make sense.

Can we just add these 10 voicebanks as an artist type field for now?

> How about making this available for editors (or only trusted or mods) to add. For example just like adding a new base/main/parent tag like the ones we have (Genres, Themes, etc.).
> We could simply add new artist "parent" types manually and the system automatically changes the colors of the "square" field.

Could you please create a new issue (or a discussion) for that?

ycanardeau:
@andreoda 

>There should be at least enough root voicebanks for an artist type field to make sense.

Can we just add these 8 voicebanks (except for chipspeech and Alter/Ego?) as an artist type for now? Does it make sense to have these new artist types considering relatively small number of entries compared to the existing ones? I guess, instead of having an `enum`, we should create a new table, if there are too many artist types.

> How about making this available for editors (or only trusted or mods) to add. For example just like adding a new base/main/parent tag like the ones we have (Genres, Themes, etc.).
[...]

PR #705, opened 2020-12-29
Add artist type for Synthesizer V

ycanardeau:
#458


VocaDbModel/Service/QueryableExtensions/SongQueryableExtensions.cs
```diff
@@ -285,6 +285,7 @@ public static IQueryable<Song> WhereHasType(this IQueryable<Song> query, SongTyp
 			SongVocalistSelection.Vocaloid => query.Where(s => s.AllArtists.Any(a => !a.IsSupport && a.Artist.ArtistType == ArtistType.Vocaloid)),
 			SongVocalistSelection.UTAU => query.Where(s => s.AllArtists.Any(a => !a.IsSupport && a.Artist.ArtistType == ArtistType.UTAU)),
 			SongVocalistSelection.CeVIO => query.Where(s => s.AllArtists.Any(a => !a.IsSupport && (a.Artist.ArtistType == ArtistType.CeVIO || a.Artist.ArtistType == ArtistType.OtherVoiceSynthesizer))),
```
ycanardeau, line 287: Why do we need `|| a.Artist.ArtistType == ArtistType.OtherVoiceSynthesizer` here?

riipah: Seems this is only used on the "top songs" page. Currently there is categories for Vocaloid, UTAU and other vocalists. Actually the "CeVIO" selection enum is reused for "other vocalists" there. Perhaps it would be best to add "Other" to SongVocalistSelection and use that instead. You could have all the artist types (including CeVIO/SynthV) there, but maybe there's not enough songs yet, so for now Vocaloid/UTAU/others would be enough? Then this SongVocalistSelection.SynthesizerV value isn't needed yet either. But I guess it can be either way.

riipah:
Looks good, but maybe consider what to do with the SongVocalistSelection still

ycanardeau:
Thank you for your review! I'll try to update it tomorrow.

Issue #723, opened 2021-01-10
API for get current user followed artist with specific artist id


up2code:
**Mobile app example usecase**
When user open artist Miku NT's detail page on VocaDB mobile app
Then app will request API to check current user with Miku NT's artist id
And API return Miku NT's follow status of current user
And app should highlighted _Like Button_  if current user followed this artist.

![qemu-system-x86_64_F6hYMFFGde](https://user-images.githubusercontent.com/2321626/104124476-aa6ca480-5383-11eb-9ee1-5c440ca2e98b.png)

**Problem**
Missing API for get followed artist with specific **artist id** by current user. 

**Possible solution**
API for get followed artist with specific **artist id** by current user or user id.

PR #725, opened 2021-01-10
Implement API for getting followed artist

ycanardeau:
#723

### DIFF
--- a/VocaDbModel/Database/Queries/UserQueries.cs
+++ b/VocaDbModel/Database/Queries/UserQueries.cs
@@ -1823,13 +1823,11 @@ namespace VocaDb.Model.Database.Queries
 				.Any(m => m.Id == messageId && m.User.Id == user.Id && m.Inbox == UserInboxType.Notifications);
 		});
 
-		public bool GetFollowedArtist(int userId, int artistId)
+		public ArtistForUserForApiContract GetArtistForUser(int userId, int artistId) => HandleQuery(ctx =>
 		{
-			if (userId == 0)
-				return false;
-
-			return HandleQuery(ctx => ctx.OfType<ArtistForUser>().Query().Any(s => s.Artist.Id == artistId && s.User.Id == userId));
-		}
+			var artistForUser = ctx.OfType<ArtistForUser>().Query().FirstOrDefault(s => s.Artist.Id == artistId && s.User.Id == userId);
+			return new ArtistForUserForApiContract(artistForUser, LanguagePreference, _entryImagePersister, ArtistOptionalFields.None);
+		});
 	}
 
 	public class AlbumTagUsageFactory : ITagUsageFactory<AlbumTagUsage>
--- a/VocaDbModel/Database/Queries/UserQueries.cs
+++ b/VocaDbModel/Database/Queries/UserQueries.cs
@@ -1822,6 +1822,14 @@ namespace VocaDb.Model.Database.Queries
 			return ctx.Query<UserMessage>()
 				.Any(m => m.Id == messageId && m.User.Id == user.Id && m.Inbox == UserInboxType.Notifications);
 		});
+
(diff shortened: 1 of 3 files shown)


VocaDbWeb/Controllers/Api/UserApiController.cs
```diff
@@ -814,5 +814,13 @@ public class PostStatusLimitedModel
...
 		[ApiExplorerSettings(IgnoreApi = true)]
 		public void PostStatusLimited(int id, [FromBody] PostStatusLimitedModel model) => _queries.SetUserToLimited(id, model.Reason, WebHelper.GetRealHost(Request), model.CreateReport);
+
+		[Route("{id:int}/followedArtists/{artistId:int}")]
+		public bool GetFollowedArtist(int id, int artistId) => _queries.GetFollowedArtist(id, artistId);
+
+		[Route("current/followedArtists/{artistId:int}")]
+		[Authorize]
+		[AuthenticatedCorsApi(System.Web.Mvc.HttpVerbs.Get)]
+		public bool GetFollowedArtist(int artistId) => GetFollowedArtist(_permissionContext.LoggedUserId, artistId);
```
ycanardeau, line 824: What should be the URL? Is it enough just to return a `bool`?

VocaDbWeb/Controllers/Api/UserApiController.cs
```diff
@@ -814,5 +814,13 @@ public class PostStatusLimitedModel
 		[Route("{id:int}/status-limited")]
 		[ApiExplorerSettings(IgnoreApi = true)]
 		public void PostStatusLimited(int id, [FromBody] PostStatusLimitedModel model) => _queries.SetUserToLimited(id, model.Reason, WebHelper.GetRealHost(Request), model.CreateReport);
+
+		[Route("{id:int}/followedArtists/{artistId:int}")]
+		public bool GetFollowedArtist(int id, int artistId) => _queries.GetFollowedArtist(id, artistId);
```
riipah, line 819: I would actually consider using ArtistForUserForApiContract for this instead of bool return value.

The reasons are: 1) you can add additional properties there such as notification settings if needed and 2) primitive values such as bool sometimes work poorly with REST APIs, wrapping inside a value is often recommended

ycanardeau: Should we return `null` if a user is not following an artist?

riipah: Or maybe ArtistForUserForApiContract with null artist?

VocaDbWeb/Controllers/Api/UserApiController.cs
riipah: The URL is fine I guess. I notice the API guidelines for VocaDB mention camelCase, although I think kebab-case is more common these days.

ycanardeau:
Thanks for your review!

VocaDbModel/Database/Queries/UserQueries.cs
```diff
@@ -1822,6 +1822,12 @@ public bool VerifyEmail(Guid requestId)
 			return ctx.Query<UserMessage>()
 				.Any(m => m.Id == messageId && m.User.Id == user.Id && m.Inbox == UserInboxType.Notifications);
 		});
+
+		public ArtistForUserForApiContract GetArtistForUser(int userId, int artistId) => HandleQuery(ctx =>
+		{
+			var artistForUser = ctx.OfType<ArtistForUser>().Query().FirstOrDefault(s => s.Artist.Id == artistId && s.User.Id == userId);
+			return new ArtistForUserForApiContract(artistForUser, LanguagePreference, _entryImagePersister, ArtistOptionalFields.None);
```
ycanardeau, line 1829: Use `ArtistOptionalFields.None` for now.

PR #727, opened 2021-01-11
Fix RegionCollection

ycanardeau:
#598

### DIFF
--- a/VocaDbModel/Domain/Globalization/RegionCollection.cs
+++ b/VocaDbModel/Domain/Globalization/RegionCollection.cs
@@ -1,5 +1,3 @@
-#nullable disable
-
 using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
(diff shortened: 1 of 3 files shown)


VocaDbModel/Domain/Globalization/RegionCollection.cs
```diff
@@ -8,27 +6,23 @@ namespace VocaDb.Model.Domain.Globalization
...
 	public class RegionCollection
 	{
-		// FIXME
-		public static readonly string[] RegionCodes = CultureInfo.GetCultures(CultureTypes.SpecificCultures)
-			.Select(culture => new RegionInfo(culture.Name).TwoLetterISORegionName)
-			.OrderBy(c => c)
-			.Distinct()
+		public static readonly IEnumerable<RegionInfo> DefaultRegions = CultureInfo.GetCultures(CultureTypes.SpecificCultures)
+			.Select(culture => new RegionInfo(culture.Name))
+			.GroupBy(region => region.TwoLetterISORegionName)
```
riipah, line 11: We also have Distinct with property selector

ycanardeau: Oh, didn't know that, thanks!

PR #736, opened 2021-01-16
Encode and parse json

ycanardeau:
We have a piece of code like this:
```javascript
<script type="text/javascript">
    $(function () {
        var jsonModel = {"name": "<script>alert('Kazemachi Hello World');</script>"};
        var latestComments = [{"message": "<script>alert('BadBye');</script>"}];
    }
</script>
```

The problem here is that the first `</script>` tag is interpreted as a closing tag. So we have actually two `<script>` blocks,
```javascript
<script type="text/javascript">
    $(function () {
        var jsonModel = {"name": "<script>alert('Kazemachi Hello World');</script>
```
and
```javascript
<script>alert('BadBye');</script>
```
That's why `alert('BadBye');` is executed. 

We have to escape the JSON somehow, but the strings generated by `HttpUtility.JavaScriptStringEncode` cannot be directly used in `<script>`:
`{\"name\": \"\u003cscript\u003ealert(\u0027Kazemachi Hello World\u0027);\u003c/script\u003e\"}` and `[{\"message\": \"\u003cscript\u003ealert(\u0027BadBye\u0027);\u003c/script\u003e\"}]`

However, this is a valid JSON string that can be parsed by `JSON.parse` in most browsers:
```javascript
var jsonModel = JSON.parse("{\"name\": \"\u003cscript\u003ealert(\u0027Kazemachi Hello World\u0027);\u003c/script\u003e\"}");
var latestComments = JSON.parse("[{\"message\": \"\u003cscript\u003ealert(\u0027BadBye\u0027);\u003c/script\u003e\"}]");
```

### DIFF
--- a/VocaDbWeb.Core/Code/VocaDbPage.cs
+++ b/VocaDbWeb.Core/Code/VocaDbPage.cs
@@ -1,6 +1,7 @@
 #nullable disable
 
 using System.Threading;
+using System.Web;
 using Microsoft.AspNetCore.Html;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.AspNetCore.Mvc.Razor;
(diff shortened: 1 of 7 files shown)


VocaDbWeb.Core/Code/VocaDbPage.cs
```diff
@@ -62,12 +63,12 @@ public string ToJS(int? val)
 
 		public IHtmlContent ToJS(string str)
 		{
-			return new HtmlString(JsonHelpers.Serialize(str));
+			return new HtmlString($@"JSON.parse(""{HttpUtility.JavaScriptStringEncode(JsonHelpers.Serialize(str))}"")");
```
riipah, line 66: You should extract this into a helper method of its own, possibly static, so it doesn't have to be repeated everywhere.

riipah:
Good that you noticed it!

It's ok to fix it like this for now, although I would extract the method so it's only in one place and can be documented properly.

If it's intended to be permanent solution then writing some unit tests for it would be good as well, but it can be merged without the tests.

ycanardeau:
Thanks for reviewing!

PR #748, opened 2021-01-22
Feat/558 add bpm field for songs

ycanardeau:


VocaDbWeb.Core/Views/Search/Partials/_SongSearchOptions.cshtml
```diff
@@ -78,6 +78,15 @@
 		</div>
 	</div>
 
+	<div class="control-group">
+		<div class="control-label">@Res.Bpm</div>
+		<div class="controls">
+			<input type="number" data-bind="textInput: minBpm" class="input-small" maxlength="10" min="20" max="400" />
```
riipah, line 84: Might be good to have the min and max BPM values as named constants somewhere

riipah:
Okay, it looks good

dset0x:
I'm a little late here! I see there's a minBpm and maxBpm field in the commit. I am not sure if these are the most useful to have.

In the vocadb discussion I see that there is usefulness in the BPM of (a) the beginning and (b) ending of the song (for mixing purposes). I also think it's good to have a (c) "Base (average) BPM". A (d) "secondary" BPM might also be useful. You can also go all out (like you mentioned in another issue) and have an arbitrarily long list of BPM segments.

While minimum and maximum can also be useful, they can also be deceiving (a 5 minute song might have a slow 10 second introduction). We could keep them, but perhaps not on their own.

Another important thing is to be able to detect whichever values we need. It would be interesting for me to see if I can put something together to detect these values and then auto-update songs.

ycanardeau:
Thanks!

PR #762, opened 2021-02-10
Replace GetCachedReport and SaveCachedReport with GetOrInsert

### LINKED ISSUE #598: Add ASP.NET Core web frontend

ycanardeau:
Closes #598.

### DIFF
--- a/VocaDbWeb.Core/Controllers/StatsController.cs
+++ b/VocaDbWeb.Core/Controllers/StatsController.cs
@@ -4,6 +4,7 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Linq.Expressions;
+using System.Runtime.Caching;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.Net.Http.Headers;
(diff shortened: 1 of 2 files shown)


VocaDbWeb.Core/Controllers/StatsController.cs
```diff
@@ -34,27 +35,20 @@ public class StatsController : ControllerBase
...
-		private T GetCachedReport<T>() where T : class => throw new NotImplementedException();
+			var item = _cache.Get($"report_{name}");
+			return item as T;
+		}
 
-		// TODO: implement
-		/*private void SaveCachedReport<T>(T data) where T : class
+		private void SaveCachedReport<T>(T data) where T : class
 		{
 			var name = ControllerContext.RouteData.Values["action"];
```
riipah, line 49: Could you move this to property, otherwise it looks fine

ycanardeau: Caching doesn't seem to work since the cache keys don't match? I replaced `GetCachedReport` and `SaveCachedReport` with `GetOrInsert` instead of moving this to a property.

ycanardeau: Caching doesn't seem to work properly since the cache keys don't match? I replaced `GetCachedReport` and `SaveCachedReport` with `GetOrInsert` instead of moving this to a property.

PR #763, opened 2021-02-10
Implement Feed and LatestVideos

### LINKED ISSUE #598: Add ASP.NET Core web frontend

ycanardeau:
Closes #598.

### DIFF
--- a/VocaDbWeb.Core/Code/Feeds/FeedResult.cs
+++ b/VocaDbWeb.Core/Code/Feeds/FeedResult.cs
@@ -1,0 +1,47 @@
+#nullable disable
+
+using System;
+using System.IO;
+using System.Net;
+using System.ServiceModel.Syndication;
+using System.Text;
+using System.Threading.Tasks;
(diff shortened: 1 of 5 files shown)


riipah:
It looks fine. This feature is not super important anyway, so if you tested that it seems to be working then it's fine.

ycanardeau:
Thanks!

PR #783, opened 2021-02-22
Enable CORS

### LINKED ISSUE #772: Unauthenticated APIs require CORS authentication
**Website**
VocaDB / UtaiteDB / TouhouDB

**Describe the bug**
Since the new update, unauthenticated APIs require CORS authentication.
The exact error is: `Access to XMLHttpRequest at '[APIRequest]' from origin '[origin]' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource.`

**Expected behavior**
As stated here: [https://github.com/VocaDB/vocadb/wiki/Public-API](https://github.com/VocaDB/vocadb/wiki/Public-API), unauthenticated API requests like  /api/songs/byPv should work without authentication.

ycanardeau:
Closes #772.

References:
- [Enable Cross-Origin Requests (CORS) in ASP.NET Core | Microsoft Docs](https://docs.microsoft.com/en-us/aspnet/core/security/cors?view=aspnetcore-5.0)
- https://github.com/aspnet/Security/issues/1569#issuecomment-350947198
- [Work with SameSite cookies in ASP.NET Core | Microsoft Docs](https://docs.microsoft.com/en-us/aspnet/core/security/samesite?view=aspnetcore-5.0)


VocaDbWeb/Startup.cs
```diff
@@ -119,6 +119,33 @@ public void ConfigureServices(IServiceCollection services)
...
+						.AllowAnyHeader()
+						.WithMethods("GET", "POST", "PUT", "OPTIONS");
+
+					var origins = AppConfig.AllowedCorsOrigins;
+					if (!string.IsNullOrEmpty(origins))
+					{
+						if (origins != "*")
+							builder.WithOrigins(origins.Split(','));
+						else
+							builder.AllowAnyOrigin();
```
riipah, line 145: Actually, for safety reasons, I would leave this out (I am aware that previously it allowed any origin if no origins are configured, but that can be really unsafe).

Also, I think that credentials (with AllowCredentials) should be enabled, because some services use the authentication cookie, such as the [UTAU lyrics wiki importer](https://github.com/VocaDB/vocadb-utau-lyrics-wiki-importer), and AllowAnyOrigin cannot be used together with AllowCredentials.

VocaDbWeb/Startup.cs
```diff
@@ -119,6 +119,33 @@ public void ConfigureServices(IServiceCollection services)
...
+						.AllowAnyOrigin()
+						.AllowAnyHeader()
+						.WithMethods("GET");
+				});
+
+				options.AddPolicy(AuthenticationConstants.AuthenticatedCorsApiPolicy, builder =>
+				{
+					builder
+						.AllowAnyHeader()
+						.WithMethods("GET", "POST", "PUT", "OPTIONS");
```
dset0x, line 137: Not at all versed in this - but is there a reason you set specific methods instead of all?

ycanardeau: Actually, we only need `GET` and `POST`.

VocaDbWeb/Startup.cs
```diff
@@ -116,9 +117,31 @@ public void ConfigureServices(IServiceCollection services)
 				.AddCookie(options =>
 				{
 					options.LoginPath = new PathString("/User/Login");
+					// See https://github.com/aspnet/Security/issues/1569#issuecomment-350947198
+					options.Cookie.SameSite = SameSiteMode.None;
```
ycanardeau, line 121: Set `SameSite` to `SameSiteMode.None`.

ycanardeau, line 121: Set `SameSite` to `SameSiteMode.None`.

VocaDbWeb/Startup.cs
ycanardeau: Was [UTAU Lyrics Wiki importer](http://vocaloid.eu/vocadb/UtauLyricsWikiImporter/) already broken? http://vocaloid.eu/utaulyricswiki/api/lyrics-wiki?pagename= throws an internal server error.

VocaDbWeb/Startup.cs
riipah: While all the PUT APIs are currently private, I see no reason why PUT would not be supported in the future. I also think it's ok to allow all verbs, including DELETE.

VocaDbWeb/Startup.cs
riipah: Could you add a comment why? Is this for external services like UTAU Wiki importer?

VocaDbWeb/Startup.cs
riipah: I don't know. I haven't used it for a while. Not sure if it's actually being used. But the point is that services like that should be supported in some way.

VocaDbWeb/Startup.cs
ycanardeau: Yes, it's for external services like UTAU Wiki importer. It's necessary to allow cross-site cookie use (e.g. from `vocaloid.eu` (in which UTAU Wiki importer is located) to `vocadb.net`).

VocaDbWeb/Startup.cs
ycanardeau: Do you mean that we can use the [AllowAnyMethod](https://github.com/dotnet/aspnetcore/blob/c925f99cddac0df90ed0bc4a07ecda6b054a0b02/src/Middleware/CORS/src/Infrastructure/CorsPolicyBuilder.cs#L181) method, which sets `Access-Control-Allow-Methods` to `*` (wildcard)?

riipah: Yeah, I see no issue with that.

ycanardeau:
Thanks for your reviews!

PR #788, opened 2021-02-24
Specify VaryByQueryKeys

### LINKED ISSUE #787: Wrong descriptions when hovering over Events
**Website**
VocaDB

**Describe the bug**
Whenever I hover over an Event it shows me the wrong description.
Same like https://github.com/VocaDB/vocadb/issues/768 but with Events

grafik

ycanardeau:
Closes #787.

### DIFF
--- a/VocaDbWeb/Controllers/StatsController.cs
+++ b/VocaDbWeb/Controllers/StatsController.cs
@@ -370,7 +370,7 @@ namespace VocaDb.Web.Controllers
 			);
 		}
 
-		[ResponseCache(Duration = ClientCacheDurationSec)]
+		[ResponseCache(Duration = ClientCacheDurationSec, VaryByQueryKeys = new[] { "cutoff" })]
 		public ActionResult ArtistsPerMonth(DateTime? cutoff = null)
 		{
 			cutoff ??= DefaultMinDate;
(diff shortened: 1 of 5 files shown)


VocaDbWeb/Controllers/StatsController.cs
```diff
@@ -601,7 +601,7 @@ public ActionResult SongsPerVocaloidOverTime(DateTime? cutoff, ArtistType[] voca
 			return AreaChart("Songs per voicebank over time", dataSeries);
 		}
 
-		[ResponseCache(Duration = ClientCacheDurationSec)]
+		[ResponseCache(Duration = ClientCacheDurationSec, VaryByQueryKeys = new[] { "cutoff", "vocalistTypes", "startYear" })]
```
riipah, line 604: I thought it was silly having to repeat all parameter names for all requests, so I tried Googling. According to Google, VaryByQueryKeys should support "*" to vary by all query keys. It's very poorly documented.

So, `VaryByQueryKeys = new[] { "*" }` should work for all of these. I think it would be better for maintainability, in addition to being shorter.

ycanardeau: I think I also read the SO answer, but does it mean the following URLs will have different caches?

- `?cutoff=3939&vocalistTypes=Vocaloid&startYear=39`
- `?cutoff=3939&vocalistTypes=Vocaloid&startYear=39&a=1`
- `?cutoff=3939&vocalistTypes=Vocaloid&startYear=39&a=1&b=2`

`[ResponseCache]` will vary by `a` and `b`, even though they are not valid query parameters for `/Stats/GetSongsPerVoicebankTypeOverTime`?

riipah: > I think I also read the SO answer, but does it mean the following URLs will have different caches? (I'm not sure if it's problematic.)

I think it will vary yes, because it seems to be reading the keys from query, not action. But I don't see it as problem. Normally you won't specify those parameters a and b. The only reason why you would add such dummy parameters is to intentionally skip (bust) the cache, which is also ok.

ycanardeau: I see, thanks for the explaination!

ycanardeau: I see, thanks for the explanation!

riipah: I should still add that maybe for more public APIs I would be more careful with using the "*", but these MVC controllers are only meant to be called from the site itself, so we control both client and server.

I'm still surprised there's no way to tell ASP.NET to vary cache by only those available parameters.

PR #791, opened 2021-02-26
#602 Text color has low contrast on Dark Angel theme

### LINKED ISSUE #602: Text color on some pages has low contrast on Dark Angel theme
VocaDB

Not a bug, but more of an accessibility issue. The contrast is too low for text in some areas of the site when using the Dark Angel theme, such as highlighted notifications and thread titles. Every time I go to the discussions page, I feel like I need to switch back to the default theme so I can actually read the thread titles. I would like to request the text color for the Dark Angel theme be changed to something a bit darker to increase contrast and improve accessibility, such as #00c7c7. Alternatively, the white/light colored background on these pages could be changed instead, if that would be better. Please consider this change.

image2
^ The discussions page using the Dark Angel theme.
image5-00c7c7
^ .rightFrame td in DarkAngel.css changed to #00c7c7

 - OS: Microsoft Windows 10 
 - Browser: Google Chrome Version 83.0.4103.106

Pyther99:
Closes #602. I also added an invert filter to the GitHub icon:
image
^Before

image
^After

### DIFF
--- a/VocaDbWeb/wwwroot/Content/Styles/DarkAngel.less
+++ b/VocaDbWeb/wwwroot/Content/Styles/DarkAngel.less
@@ -1,4 +1,4 @@
-﻿@light-green: rgb(150, 220, 220);
+@light-green: rgb(150, 220, 220);
 @medium-green: #64B7B7;
 
 body {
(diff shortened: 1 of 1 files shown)


ycanardeau (approved):
Thank you! By the way, you don't need to use the future branch.

Issue #299, opened 2017-07-05
Artist role for original performer


riipah:
Currently indicated with tags. For example [Kyary Pamyu Pamyu](http://vocadb.net/T/3234/kyary-pamyu-pamyu). An artist link could be more appropriate, but a new role is needed.

Benefits:
* Artist type can be specified
* Artist can be tagged

Drawbacks:
* Need to customize roles

Issue #737, opened 2021-01-16
New artist type for cover artists


andreoda:
**Is your feature request related to a problem? Please describe.**
Some users/visitors are amused or mad by the fact that they are marked as "producer" on VocaDB even though they only produce covers with voice synths, not actual instrumentals/music. It leads to confusion.

**Describe the solution you'd like**
A new artist type for people who only produces covers. Example: "cover artist"

They should only get the "producer" type if they at least produced one song featuring their own composition/remix. Similar to the rule on the Discord server for the "producer" role.

**Describe alternatives you've considered**
-> https://github.com/VocaDB/vocadb/issues/620 (Show artist subtype tag on details page and tooltips)

**Additional context**
Related issues:
-> https://github.com/VocaDB/vocadb/issues/413 (Suggest remix and cover tags for artists based on song types)
-> https://github.com/VocaDB/vocadb/issues/704 (Making it possible to create new artist types on VocaDB) - should be considered first

Issue #620, opened 2020-09-25
Artists: Show artist subtype tag on details page and tooltips


riipah:
I have a feeling there might be an issue for this already, but couldn't find.

We have artist tags such as Vocaloid engine version (Vocaloid 2, Vocaloid 3 etc.) and other voice synthesizer (Synthesizer V etc.). It would be useful to show those tags on artist details page next to artist type, and in tooltips, so that user would see that a voicebank is for example Vocaloid 5 engine (instead of just Vocaloid). This would be even more useful for other voice synthesizer engines without their own type. Instead of the "other voice synthesizer" artist type, we could even just show the subtype (Synthesizer V).

Do we need a new flag for artist entries? Or is it enough to scan the child tags of certain voicebank tags (Vocaloid and OtherVoiceSynthesizer)? 

Could we also consider including language tags? So that it could show "Vocaloid (English)" as the artist type.

For performance reasons the associated tag could be cached directly on the artist entity. But what to do with multiple tags? Maybe caching isn't needed right away.

If we add a flag, it could have two values:
* Show as subtype (show main type + subtype, for example Vocaloid (Vocaloid 2))
* Replace main type (do not show main type)

Issue #326, opened 2017-10-04
New artist role for subject


riipah:
If we decide to keep the artbooks, it would be useful to be able to connect characters featured in the artbooks with the "subject" role.

* Should this role only apply to artbooks? It would be useful for animation PVs as well.
* Should this be the default role for Vocaloids in artbooks? Currently it's not possible to customize the roles for Vocaloids. Do we need other roles?
* Could the subject role be used for character designs as well, for example Hagane Miku? See #519.

ycanardeau:
[mentioned](https://discord.com/channels/309072240639737866/666326527197315091/852273096676081694)

PR #792, opened 2021-02-26
New Artist Type DJ and Artist Role original performer

### LINKED ISSUE #319: Artist type for DJ

Pyther99:
Closes #319 and #299.

**DJ**
-> ArtistCategory Producer to avoid the need of role specification in remixes
-> Customizable to allow the role of mixer in Producer DJ collabs

**Original performer**
-> Example. [Kyary Pamyu Pamyu](https://vocadb.net/T/3234/kyary-pamyu-pamyu)

VocaDb wiki needs to be updated:
[Artist Types](https://wiki.vocadb.net/SiteSettings/Wiki/Index/23?title=Artist-types)
[Artist Roles](https://wiki.vocadb.net/SiteSettings/Wiki/Index/21?title=Artist-roles)

### DIFF
--- a/VocaDbModel/Domain/Artists/ArtistRoles.cs
+++ b/VocaDbModel/Domain/Artists/ArtistRoles.cs
@@ -73,5 +73,7 @@ namespace VocaDb.Model.Domain.Artists
 		Encoder = 1 << 14,
 
 		VocalDataProvider = 1 << 15,
+
+		OriginalPerformer = 1 << 16,
 	}
 }
--- a/VocaDbModel/Domain/Artists/ArtistType.cs
+++ b/VocaDbModel/Domain/Artists/ArtistType.cs
@@ -59,6 +59,8 @@ namespace VocaDb.Model.Domain.Artists
 		Character,
 
 		SynthesizerV,
+
+		DJ,
 	}
 
 	/// <summary>
@@ -105,5 +107,7 @@ namespace VocaDb.Model.Domain.Artists
 		Character = 1 << 16,
 
 		SynthesizerV = 1 << 17,
+
+		DJ = 1 << 18,
 	}
 }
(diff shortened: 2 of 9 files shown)


ycanardeau:
As you can see, the current implementation allows up to 32 artist types. That's why I'm reluctant to add a new artist type and why I only added the one for `SynthesizerV` in #705. It's better not to add a new artist type as long as it can be replaced by a tag. #704 would be a better solution for this, which needs planning. Perhaps we should discuss this topic on Discord.

ycanardeau:
It's better not to add a new artist type as long as it can be replaced by a tag. #704 would be a better solution, which requires more planning. Perhaps we should discuss this topic on Discord. I'm reluctant to add a new artist type and that's why I only added the one for `SynthesizerV` in #705.

See also:
- https://discord.com/channels/309072240639737866/666326527197315091/797839790938521670
- https://discord.com/channels/309072240639737866/666326527197315091/797839909061918800

Pyther99:
Sure, I just don't think there are enough artist types to make #704 worthwhile. The currently available (and suggested) roles already cover the majority of artists. Besides, too many types could confuse new users (DJ is relatively self-explanatory).

List of suggested but not implemented types and roles:
-> DJ (#319)
-> Cover artist (#737)
-> Original performer (#299)
-> subject (#326)

Furthermore, if #620 is implemented, users could add subtypes to artists with tags, making #704 redundant in my opinion.

ycanardeau:
Did you see https://github.com/VocaDB/vocadb/issues/458#issuecomment-732857985?

Pyther99:
Ah, I forgot about that comment. But then I would also suggest redesigning the "Add an Artist" form so that the drop-down menu is not cluttered with artist types. (Maybe as a part of #305).

PR #806, opened 2021-03-09
Update song length from PV

### LINKED ISSUE #803: "Refresh PV metadata" wont't update/add the song length
Maintenance action "Refresh PV metadata" does not update/add the PV length. 

This fix would be useful with updating the song entries that have the incorrect length "0:00", mostly from bilibili (around 1300 songs).

Example entries:

https://vocadb.net/S/1419
https://vocadb.net/S/193284

ycanardeau:
Closes #803.

### DIFF
--- a/VocaDbModel/Domain/Songs/PVForSong.cs
+++ b/VocaDbModel/Domain/Songs/PVForSong.cs
@@ -119,6 +119,7 @@ namespace VocaDb.Model.Domain.Songs
 			Author = result.Author;
 			ExtendedMetadata = result.ExtendedMetadata;
 			ThumbUrl = result.ThumbUrl;
+			Length = result.LengthSeconds ?? 0;
 		}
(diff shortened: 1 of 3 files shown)


VocaDbModel/Domain/Songs/PVForSong.cs
```diff
@@ -119,6 +119,7 @@ public virtual async Task RefreshMetadata(IPVParser pvParser, IUserPermissionCon
 			Author = result.Author;
 			ExtendedMetadata = result.ExtendedMetadata;
 			ThumbUrl = result.ThumbUrl;
+			Length = result.LengthSeconds ?? 0;
```
riipah, line 122: I think it should default to current length (this.Length), if for some reason we could not get length.

ycanardeau:
Thanks!

PR #822, opened 2021-04-03
Install Swashbuckle.AspNetCore.Newtonsoft

### LINKED ISSUE #814: Swagger API has type problems
This concerns https://vocadb.net/swagger/index.html and https://vocadb.net/swagger/v1/swagger.json

1. Many parameters are shown as numbers. For example, in /api/songs/byPv, pvService is exposed as an int32.
2. Some queries may return null, but this is not reflected in the json. For example, /api/songs/byPv claims to _always_ return #/components/schemas/SongForApiContract which is a non-null object. However, null is returned if the PV does not exist.

ycanardeau:
Fixes #814.


riipah:
Looks fine...

PR #823, opened 2021-04-03
added mailmap

riipah:
Added git mailmap to get rid of some email aliases. Although I'm not sure if Github supports it, at least local git history will be more clear.


riipah:
@ycanardeau can you check?

ycanardeau:
Looks good, thanks!

PR #836, opened 2021-04-12
Autofill label from URL when adding lyrics to a song

### LINKED ISSUE #833: Autofill label from URL when adding lyrics to a song
When adding lyrics to a song entry, it would be convenient if the label was automatically filled when pasting source URLs (piapro in most cases), similarly to the song external links.

Pyther99:
Closes #833.

### DIFF
--- a/VocaDbWeb/wwwroot/Scripts/ViewModels/Song/LyricsForSongEditViewModel.ts
+++ b/VocaDbWeb/wwwroot/Scripts/ViewModels/Song/LyricsForSongEditViewModel.ts
@@ -31,11 +31,11 @@ import WebLinkMatcher from '../../Shared/WebLinkMatcher';
 				if (!this.source()) {
 
 					var matcher = WebLinkMatcher.matchWebLink(url);
-				}
 
-				if (matcher) {
+					if (matcher) {
 
-					this.source(matcher.desc);
+						this.source(matcher.desc);
+					}
 				}
 
 			});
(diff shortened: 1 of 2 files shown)


VocaDbWeb/wwwroot/Scripts/ViewModels/Song/LyricsForSongEditViewModel.ts
```diff
@@ -25,6 +26,20 @@ import TranslationType from '../../Models/Globalization/TranslationType';
...
+
+				if (!this.source()) {
+
+					var matcher = WebLinkMatcher.matchWebLink(url);
+				}
+
+				if (matcher) {
+
+					this.source(matcher.desc);
+				}
```
ycanardeau, line 39: Is there any reason to put this outside the `if (!this.source())` block?

ycanardeau:
https://github.com/VocaDB/vocadb/blob/f1ef5e040e82b0ba91f28531d4936542f856af2e/VocaDbWeb/Views/Song/Partials/_LyricsForSongEdit.cshtml#L29

@Pyther99 Which do you think is better, `data-bind="textInput: source"` or `data-bind="value: source"`?

Pyther99:
@ycanardeau I would prefer using `textInput`, because it handles unusual text entry methods better. ([Knockout Documentation](https://knockoutjs.com/documentation/textinput-binding.html#note-1-textinput-vs-value-binding)). 
Other link auto completion fields seem to use the `value`-Binding https://github.com/VocaDB/vocadb/blob/f1ef5e040e82b0ba91f28531d4936542f856af2e/VocaDbWeb/Views/Shared/Partials/Knockout/_WebLinksEditViewKnockout.cshtml#L29
so maybe `_WebLinksEditViewKnockout.cshtml` should get changed to `textInput` for consistency.

ycanardeau:
For consistency, if possible, would it be better to replace `textInput` in [_LyricsForSongEdit.cshtml](https://github.com/VocaDB/vocadb/blob/f1ef5e040e82b0ba91f28531d4936542f856af2e/VocaDbWeb/Views/Song/Partials/_LyricsForSongEdit.cshtml) with `value` for now, rather than to replace `value` with `textInput` in [_WebLinksEditViewKnockout.cshtml](https://github.com/VocaDB/vocadb/blob/6182c2174689c6129f9f0b2c2bb265aac1149725/VocaDbWeb/Views/Shared/Partials/Knockout/_WebLinksEditViewKnockout.cshtml)? Perhaps we could get a proper fix later in the future.

ycanardeau:
@Pyther99 I've replaced `textInput` with `value`.

PR #820, opened 2021-03-31
Song Creation form is not sufficiently user friendly #305

Pyther99:
Depends on #817.
Closes #305.
Closes #14.
Partially implements #368. (Romanization link)

Implementation of @dset0x's patch for the song creation page. 

- [x] This PR needs to be merged after #817 to include the new alert
- [x]  The current alert texts need to be split 
- [x] The new alert texts need to be translated

Needs to be tested on the beta site.

PR #839, opened 2021-04-15
Alerts in one column

Pyther99:
Fixes various bugs:
- Decrease fieldset margin
- Change "Vocaloid" to "Voice synthesizer" in SongRes.resx
- Move alerts into one column

### DIFF
--- a/VocaDbWeb/wwwroot/Content/Styles/Overrides.less
+++ b/VocaDbWeb/wwwroot/Content/Styles/Overrides.less
@@ -222,8 +222,34 @@ ul.nav.nav-pills {
 	margin-bottom: 0;
 }
 
+legend {
+	margin-bottom: 0;
+	border: none;
+	line-height: 1em;
+}
+
 fieldset {
-	margin: 0 0 0;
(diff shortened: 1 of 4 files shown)


Pyther99:
localhost_44398_Song_Create (4)

VocaDbWeb/wwwroot/Content/Styles/Overrides.less
```diff
@@ -222,18 +222,48 @@ ul.nav.nav-pills {
...
+	font-weight: 500;
+	font-size: 13px;
+	float: right;
+	padding: 4px 7px;
+}
+
+.alert {
+	margin-bottom: 0;
+	padding: 8px 14px;
+}
```
ycanardeau, line 248: It seems that this causes unwanted side effects. For example, compare https://vocadb.net/Album/Create and https://beta.vocadb.net/Album/Create. I'm not sure if there would be any unwanted side effects other than this. It's surprisingly difficult to style HTML with CSS. I think it's better not to merge #820 and #839 into main for now.

Issue #95, opened 2015-10-11
Account / User profile / Front page: Customizable front page


riipah:
Logged in users could be able to customize the front page, mostly by disabling elements they don't need. For example, old users might not find the "random popular albums" list that useful.

PR #850, opened 2021-04-26
Increase the amount of visible events

### LINKED ISSUE #849: Increase the amount of visible events on the homepage
A large event is currently hidden from the home page due to other events taking precedence because they're more recent.

https://vocadb.net/E/3784/the-vocaloid-collection-2021-s

Currently, the limit seems to be 3: https://github.com/VocaDB/vocadb/blob/821e292b118e2049e44bb1474e5db002f8c374dc/VocaDbModel/Service/OtherService.cs

Maybe 4-6 would be better.

Pyther99:
Closes #849
I reduced the width of `.event-teaser` to keep events in one row. More events could cause bad line wrapping so everything beyond 4 needs a new row (Maybe implement that as an option with #95 ).

### DIFF
--- a/VocaDbModel/Service/OtherService.cs
+++ b/VocaDbModel/Service/OtherService.cs
@@ -149,7 +149,7 @@ namespace VocaDb.Model.Service
 
 		private ReleaseEventForApiContract[] GetRecentEvents(ISession session)
 		{
-			var count = 3;
+			var count = 4;
 			var cacheKey = $"OtherService.RecentEvents.{LanguagePreference}";
 			return _cache.GetOrInsert(cacheKey, CachePolicy.AbsoluteExpiration(24), () =>
 			{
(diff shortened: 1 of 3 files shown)


Pyther99:
image

ycanardeau:
@Pyther99 What do you mean by "bad line wrapping"?

Pyther99:
@ycanardeau 
image

ycanardeau:
@Pyther99 I see, thanks!

@Shiroizu Deployed on the beta site.

ycanardeau:
@Pyther99 It seems that this causes bad line wrapping when shrinking the width of the browser window.

image

Pyther99:
@ycanardeau That can also happen on the current site and should be adressed in #6. Just with more than 4 events this happens without shrinking the width of the window.

ycanardeau:
@Pyther99 Then, let's do this as part of #6 in the future. It's not a good idea to spend much time on reviewing, testing and fixing legacy code.

ycanardeau:
@Pyther99 #900 has been taking much more time than I expected, so I'll merge this now. Thanks for the PR!

PR #874, opened 2021-05-15
Discord webhook

ycanardeau:


VocaDbModel/Utils/AppConfig.cs
```diff
@@ -239,5 +239,9 @@ public static SongType[] SongTypes
 		public static string YoutubeApiKey => Val("YoutubeApiKey");
 
 		public static string VimeoApiKey => Val("VimeoApiKey");
+
+		public static string DiscordWebhookUrl = Val("DiscordWebhookUrl");
```
riipah, line 243: Might be better to use "=>" here, same as others, so we don't need to read this for tests.

riipah: And maybe add some comments here explaining the purpose

ycanardeau: Oh, thanks. This is due to my new keyboard.

ycanardeau:
@riipah I moved the code to `DiscordWebhookNotifier` because I realized that sending notifications should be optional.

PR #893, opened 2021-06-16
Include albums whose release dates only specified by year/month in sorted queries

### LINKED ISSUE #876: bilibili embedded window needs adjusting
**Website**
VocaDB

**Describe the bug**
The bilibili embedded window needs adjusting in size, the bottom shows a bit of unnecessary space. That space also shows a corner of a popup window that can be accidentally clicked on and leads to another page (Commercial for the bilibili app). It is pretty annoying when actually trying to click on the play button.

window

### LINKED ISSUE #881: Include albums whose release dates were specified only by year/month in sorted queries
**Is your feature request related to a problem? Please describe.**
When querying albums, if one chooses "sort by release date", only albums whose release date have been specified by date are listed. Albums with release date specified like "2020/10" won't show up in the query result. This can be confusing (if not misleading) to readers.

There're many albums who have no accurate release date announced; producers just put up the ordering some day, and start shipping whatever else day when everything gets ready to ship. And our editors would finalize the release date specified to the month people receives the album, leaving the date field blank.

**Describe the solution you'd like**
Currently the logic is 
```
public static IQueryable<Album> WhereHasReleaseDate(this IQueryable<Album> criteria) 
{ 
    return criteria.Where(a => a.OriginalRelease.ReleaseDate.Year != null 
        && a.OriginalRelease.ReleaseDate.Month != null && a.OriginalRelease.ReleaseDate.Day != null); 
}
```
https://github.com/VocaDB/vocadb/blob/585136f696bdff161acbd3c6e6599860a66d8386/VocaDbModel/Service/QueryableExtensions/AlbumQueryableExtensions.cs#L133

And we may change this behaviour.

AgFlore:
Closes #881.
Closes #876.

There're still debateable questions around this issue, for example, is the sorted-query meant to be a full list of all albums available or not?

If yes, then we shall allow all albums in (just like the cases for songs), not just those who has only release date given. If an album is estimated to be released in "2021", then we shall let it show up somewhere around 2021 in the list.

If no, then this has been sufficient, except that maybe we need to notify our readers on the webpage that this is not an complete list.


ycanardeau:
Available on the [beta site](http://beta.vocadb.net/). I think it doesn't have to be a full list of all albums available, but you can discuss about it on Discord.

AgFlore:
Seems functioning properly for me (except that the page got stuck at my first search, probably due to some unfinished fresh-start indexing stuff).

So here's the decisions to be made:
1. Shall we allow albums that only have release date specified to "year" in the sorted query result? Technically there would be nothing difficult or unnatural with this, it's just like what we have done for the month-albums. Albums released on "2020" will show up in the list after those released on "2020/01/01" or "2020/01" and so on.
2. Under the current mechanism, basically not all albums show up in the release-date-sorted result. If an album does not have a release date specified to date/to month/to year (well, all albums are required to have a release year, don't they?), then it will show up in the name-sort/ranking-sort, but not the release-date-sort, which may be anti-intuitive and even confusing for readers. We may either add a prompt in the page saying "if you choose to sort with release date then you are being returned with an incomplete list", or we may just abolish the whole filtering thing, and return every album in the sorted result, leaving albums that have no release year to the bottom of the sorted list.

Personally i'm a bit into the totally-abolish-the-filter option, so that editors can easily sort out which albums are having their release date completely missing. Will turn to discord to talk about that.

AgFlore:
@riipah can you remember why was the filtering introduced in the first place? Is there any special significance in filtering out albums that have incomplete release dates? (and is it safe to abolish it?)

riipah:
Maybe the only reason to keep the filtering is that it might look confusing if there's a lot of "zero values" at the end. I don't think there is any other reason. Probably when the filter was added there was more albums without release date.

AgFlore:
> Maybe the only reason to keep the filtering is that it might look confusing if there's a lot of "zero values" at the end. I don't think there is any other reason. Probably when the filter was added there was more albums without release date.

I see. So I'm amending most of the filtering so that they only require the release year to present.

Seems that there's an `ReleaseDateWithNulls` sort rule, and (though it seems that it's not accessible from the web interface) can be accessed from the api, so i'm not implementing the unfiltered search at least in this commit.

VocaDbModel/Service/Search/AlbumSearch/AlbumSearchSort.cs
```diff
@@ -11,9 +11,7 @@ public class AlbumSearchSort
 	{
 		private static IQueryable<Album> AddReleaseRestriction(IQueryable<Album> criteria)
 		{
-			return criteria.Where(a => a.OriginalRelease.ReleaseDate.Year != null
-				&& a.OriginalRelease.ReleaseDate.Month != null
-				&& a.OriginalRelease.ReleaseDate.Day != null);
+			return criteria.Where(a => a.OriginalRelease.ReleaseDate.Year != null);
 		}
```
ycanardeau, line 15: This method is only used by https://github.com/VocaDB/vocadb/blob/465f446275922de04689739b33ade382bed8a3f6/VocaDbModel/Service/Search/AlbumSearch/AlbumSearchSort.cs#L17
which is only used by https://github.com/VocaDB/vocadb/blob/465f446275922de04689739b33ade382bed8a3f6/VocaDbModel/Service/AlbumService.cs#L52
which is only used by https://github.com/VocaDB/vocadb/blob/465f446275922de04689739b33ade382bed8a3f6/VocaDbModel/Service/AlbumService.cs#L45
which is only used by https://github.com/VocaDB/vocadb/blob/465f446275922de04689739b33ade382bed8a3f6/VocaDbModel/Service/AlbumService.cs#L124
which is not used. Can we remove all of these methods?

AgFlore: Resulting in the whole deletion of the `vocadb/VocaDbModel/Service/Search/AlbumSearch/AlbumSearchSort.cs` file, if I'm understanding correctly?

ycanardeau: Yes.

ycanardeau:
Can you please add unit tests for `AlbumForUserQueryableExtensions.WhereHasReleaseYear` and `AlbumQueryableExtensions.WhereHasReleaseYear` [here](https://github.com/VocaDB/vocadb/tree/402faf0ea7eaf45d2f14cc04bcad69cba1f5da9d/Tests/Service)?

By the way, https://github.com/VocaDB/vocadb/wiki/Architecture helped me a lot to understand the big picture of the structure.

ycanardeau:
Thanks for the PR!

AgFlore:
Ahh thanks for your help! Was a bit struggling with understanding c sharps so sorry for the delay >︿<

Issue #900, opened 2021-07-03
React conversion


ycanardeau:
## Tasks

- [x] #862
- [x] #901
- [x] #1133
- [x] #902
- [x] #906
- [x] #909
- [x] #1000
- [ ] #1067
- [ ] #853
- [ ] #1169
- [x] #1170

## Motivation

Currently we are facing several issues with frontend development. 

It's impossible to find JavaScript errors in cshtml files at compile-time, rather than at run-time.

JavaScript code is scattered around ts and cshtml files, which makes code splitting and dynamic imports difficult. Now that we are using Webpack through Laravel Mix (#5 and #540, powered by [ReMikus](https://github.com/VocaDB/ReMikus)), we can easily make use of [code splitting in React](https://reactjs.org/docs/code-splitting.html). See also #872.

The jQuery and jQuery UI packages cannot be upgraded easily, because there seem to be some breaking changes. This is why #675 and #679 cannot be merged.

It requires boilerplate code to translate user interfaces. In the current implementation, we need to
1. create/edit a resx file
1. update ResourcesApiController if needed
1. create a ResourceRepository and pass it to a viewmodel
1. create a ResourcesManager with the repository and current UI culture.
1. load resources

, which is a lot of pain. By using [i18next](https://github.com/i18next/i18next) and [react-i18next](https://github.com/i18next/react-i18next), we can make this process a lot easier.

We have two different libraries for the same thing, for example marked and MarkdownSharp (#38 and #904).

## Why React?

There are several reasons why I chose React, but the main reasons are that
1. React is developed and maintained by Facebook and a community of individual developers and companies.
1. React is used by [DeviantArt](https://www.deviantart.com/), [Discord](https://discord.com/), [Mastodon](https://joinmastodon.org/) and [MusicBrainz](https://musicbrainz.org/).
1. TypeScript supports JSX.
1. I like the React way of thinking (one-way data flow, function components, JSX and etc.).
1. There is a way to [integrate with other libraries](https://reactjs.org/docs/integrating-with-other-libraries.html).
1. Using create-react-app is optional.
1. We are still using customized [Bootstrap 2](https://getbootstrap.com/2.3.2/components.html#alerts) and this could be replaced with [react-bootstrap](https://github.com/react-bootstrap/react-bootstrap).

[...]

PR #901, opened 2021-07-03
Install packages for React, create ReactController

ycanardeau:
First step towards #900.


VocaDbWeb/Startup.cs
```diff
@@ -370,6 +370,8 @@ void HandleHttpError(int code, string? description = null, string? msg = null)
 				endpoints.MapControllerRoute(
 					name: "default",
 					pattern: "{controller=Home}/{action=Index}/{id?}");
+
+				endpoints.MapFallbackToController(action: "Index", controller: "React");
```
ycanardeau, line 374: The `ReactController` controller will handle requests to the React app, which will simply serve the single page in the React app.

PR #919, opened 2021-07-13
Create DiscussionIndexStore

ycanardeau:


VocaDbWeb/tsconfig.json
```diff
@@ -25,7 +25,8 @@
 		},
 		"resolveJsonModule": true,
 		"experimentalDecorators": true,
-		"emitDecoratorMetadata": true
+		"emitDecoratorMetadata": true,
+		"useDefineForClassFields": true
```
ycanardeau, line 29: See https://github.com/mobxjs/mobx/issues/2843#issuecomment-794137196 and [Migrating from MobX 4/5 · MobX](https://mobx.js.org/migrating-from-4-or-5.html).
> Add the flag `"useDefineForClassFields": true` to your compiler config.

VocaDbWeb/Scripts/Stores/Discussion/DiscussionIndexStore.ts
```diff
@@ -0,0 +1,185 @@
...
+		reaction(
+			() => this.selectedFolder,
+			(folder) => {
+				this.showCreateNewTopic = false;
+				this.selectedTopic = undefined;
+				this.paging.goToFirstPage();
+
+				this.loadTopics(folder);
+			},
+		);
```
ycanardeau, line 84: The callback of `reaction` is *not* called if the new value is equal to the old value, while the callback of `subscribe` is called. We need to do something if we want the exact same behavior with Knockout's subscribe.

Let's say topic 39 is located in the folder 1.

| Knockout | MobX |
| --- | --- |
| Go to `/discussion/topics/39`. | Go to `/discussion/topics/39`. |
| The `selectedFolder` is now folder 1. | The `selectedFolder` is now folder 1. |
| Click on the folder name on the breadcrumb. | Click on the folder name on the breadcrumb |
| The route has changed, and `selectFolderById` is called. | The route has changed, and `selectFolderById` is called. |
| The `selectedFolder` is set to folder 1 (unchanged). | The `selectedFolder` is set to folder 1 (unchanged). |
| The callback of `subscribe` is called. (old: folder 1, new: folder 1) | The callback of `reaction` is *not* called. |

PR #938, opened 2021-08-08
Update the integration with SoundCloud API

### LINKED ISSUE #927: SoundCloud API changes
Email from SoundCloud:

----
Dear partner / API user,

If you are not calling `api.soundcloud.com` you can stop reading right here. Otherwise, there are a number of updates that might require actions from your side.

At the beginning of July, we announced security updates to our API, mainly how we deal with auth. Please read a blog post to know more about the changes.

We understand that it’s a lot and might require some time on your side to update the systems. We would appreciate it if you could start working on it ASAP, as some of the changes are live already.

Deprecation of `client_id` and the requirement to provide an `Authorization` header for each request is the biggest update. We are planning to start slowly rolling it out in September. Meaning that some percentage of requests will start failing if the header is not present.

If you have no capacity to update your systems in the following 1-2 months please let us know. We will try to find a solution. We definitely do not want to create a bad experience for users.

For any further questions please do not hesitate to contact us. You can comment on this thread, or create a new ticket.

Best,

Soundcloud Team 

----

This does concern VocaDB, and sounds like it needs changes.

Also documented here: https://developers.soundcloud.com/blog/security-updates-api

ycanardeau:
Fixes #927.

### DIFF
--- a/VocaDbModel/Service/VideoServices/VideoServiceSoundCloud.cs
+++ b/VocaDbModel/Service/VideoServices/VideoServiceSoundCloud.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Net;
 using System.Net.Http;
+using System.Net.Http.Headers;
 using System.Threading.Tasks;
 using Newtonsoft.Json;
 using NLog;
@@ -15,6 +16,11 @@ namespace VocaDb.Model.Service.VideoServices
 {
 	public class VideoServiceSoundCloud : VideoService
 	{
+		private sealed record SoundCloudTokenResponse
+		{
+			public string Access_token { get; init; } = default!;
+		}
+
 		private sealed record SoundCloudUser
 		{
 			public string Avatar_url { get; init; } = default!;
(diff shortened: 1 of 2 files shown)


VocaDbModel/Service/VideoServices/VideoServiceSoundCloud.cs
```diff
@@ -73,7 +77,25 @@ VideoUrlParseResult ReturnError(Exception x, string additionalInfo = null)
...
+				{
+					new("client_id", apikey),
+					new("client_secret", AppConfig.SoundCloudClientSecret),
+					new("grant_type", "client_credentials"),
+				}));
+				tokenResponse.EnsureSuccessStatusCode();
+				var text = await tokenResponse.Content.ReadAsStringAsync();
+
+				var response = JsonConvert.DeserializeObject<SoundCloudTokenResponse>(text);
+				var accessToken = response.Access_token;
```
ycanardeau, line 93: We should check if `response` is null, since `JsonConvert.DeserializeObject` may return null. Should we throw an exception in that case?

VocaDbModel/Service/VideoServices/VideoServiceSoundCloud.cs
```diff
@@ -73,7 +77,25 @@ VideoUrlParseResult ReturnError(Exception x, string additionalInfo = null)
...
+
+				result = await JsonRequest.ReadObjectAsync<SoundCloudResult>(apiUrl, timeout: TimeSpan.FromSeconds(10), headers: headers =>
+				{
+					headers.Authorization = new AuthenticationHeaderValue("OAuth", accessToken);
+				});
```
ycanardeau, line 98: We should check if `result` is null, otherwise a `NullPointerException` will be thrown later in this code.

ycanardeau, line 98: Anyway, `response` and `result` should never be null except for API changes.

PR #978, opened 2021-10-03
Set default value as empty string

### LINKED ISSUE #976: Error reverting song entry to an earlier version
**Website**
VocaDB

**Describe the bug**
Reverting a song entry to an earlier version causes an error.

**To Reproduce**
1. Go to https://vocadb.net/Song/ViewVersion/1083860
2. Click "Revert to this version", confirm
3. Error

**Additional context**
User that closed the report replaced lyrics and romanization that I corrected and proofread with inaccurate ones, and I have no backups.

*EDIT: the issue with lyrics has been resolved, there is no need to revert the particular entry from my example.*

ycanardeau:
Closes #976.


ycanardeau:
Let's do this again.

PR #1063, opened 2022-02-08
Feat/1062 required notes field for broken PV

### LINKED ISSUE #1062: "Notes"-field for "Broken PV" -entry reports to be required before sending.
False "Broken PV" entry reports are weirdly common. I suspect that it's just people randomly clicking the "Report an error" -button (Broken PV being the default report type). Usually the PV still works despite the report.

Having the "notes"-field be a required field before sending the report would help filtering out the false reports. Having it required for all the entry report types probably wouldn't hurt (invalid or missing information, duplicate entry, etc.).

ycanardeau:
Closes #1062.

### DIFF
--- a/Tests/GitHubIssues/Issue1062/ArtistQueriesTests.cs
+++ b/Tests/GitHubIssues/Issue1062/ArtistQueriesTests.cs
@@ -1,0 +1,144 @@
+using System.Runtime.Caching;
+using System.Threading.Tasks;
+using FluentAssertions;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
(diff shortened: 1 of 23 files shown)


ycanardeau:
@Shiroizu This is now available on the beta site. And I've implemented a feature where an error message will be shown if an entry report cannot be sent.

Shiroizu:
@[ycanardeau](https://github.com/ycanardeau) Thanks! Tested a few song entries and it seemed to work properly.

ycanardeau:
@Shiroizu Thanks for testing it!

PR #1094, opened 2022-03-25
Improved and updated Spanish translation #3

FranGamer1892:
I am starting to go through all resx files checking if the Spanish translation is in order. This is what I got until now.


FranGamer1892:
Upon further examination, I believe that these commits can be pushed safely. Of course, I'll try working on more resx files as time goes on. Unfortunately, I don't seem to be getting any responses from other Spanish-speaking contributors so they can check my work.

Untitled39:
Checked all new changes, everything seems good. 👍

FranGamer1892:
Thank you!

ycanardeau:
Thanks for the PR and review!

PR #1129, opened 2022-05-03
increased description length for event series to 4000

kateabr:


ycanardeau:
You need to increase not only the NHibernate mapping, but also the [`maxlength`](https://github.com/VocaDB/vocadb/blob/c6c18a89d20063cd1a729cc7ad013df82fab3ac1/VocaDbWeb/Views/Event/EditSeries.cshtml#L86) property of the `textarea` tag.

ycanardeau:
Thanks!

Issue #155, opened 2016-06-14
Allow uploading custom thumbnails for PVs


riipah:
At least for those hosted on VocaDB (localfile). Maybe for piapro as well.

PR #1199, opened 2022-08-16
Added a new event enum value, "Festival"

kateabr:
Closes VocaDB/community#155.

### DIFF
--- a/Tests/Domain/ReleaseEvents/ReleaseEventTypesTests.cs
+++ b/Tests/Domain/ReleaseEvents/ReleaseEventTypesTests.cs
@@ -12,7 +12,8 @@ public class ArtistEventTypesTests
 	[DataRow(8, EventCategory.Concert)]
 	[DataRow(16, EventCategory.Contest)]
 	[DataRow(32, EventCategory.Convention)]
-	[DataRow(64, EventCategory.Festival)]
+	[DataRow(64, EventCategory.Other)]
+	[DataRow(128, EventCategory.Festival)]
 	[TestMethod]
 	public void Value(int expected, EventCategory actual)
 	{
--- a/Tests/Domain/ReleaseEvents/ReleaseEventTypesTests.cs
+++ b/Tests/Domain/ReleaseEvents/ReleaseEventTypesTests.cs
@@ -5,18 +5,18 @@ namespace VocaDb.Tests.Domain.ReleaseEvents;
 [TestClass]
 public class ArtistEventTypesTests
 {
-	[DataRow(0, EventCategory.Unspecified)]
-	[DataRow(1, EventCategory.AlbumRelease)]
-	[DataRow(2, EventCategory.Anniversary)]
-	[DataRow(4, EventCategory.Club)]
-	[DataRow(8, EventCategory.Concert)]
-	[DataRow(16, EventCategory.Contest)]
-	[DataRow(32, EventCategory.Convention)]
-	[DataRow(64, EventCategory.Other)]
-	[DataRow(128, EventCategory.Festival)]
+	[DataRow("Unspecified", nameof(EventCategory.Unspecified))]
+	[DataRow("AlbumRelease", nameof(EventCategory.AlbumRelease))]
+	[DataRow("Anniversary", nameof(EventCategory.Anniversary))]
+	[DataRow("Club", nameof(EventCategory.Club))]
+	[DataRow("Concert", nameof(EventCategory.Concert))]
+	[DataRow("Contest", nameof(EventCategory.Contest))]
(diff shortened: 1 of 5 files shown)


kateabr:
#155

VocaDbWeb/wwwroot/locales/en/VocaDb.Web.Resources.Domain.ReleaseEvents.json
```diff
@@ -18,8 +18,9 @@
     "Club": "Club",
     "Concert": "Concert",
     "Contest": "Contest",
-    "Convention": "Convention",
-    "Other": "Other",
+	"Festival": "Festival",
```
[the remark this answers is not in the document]
kateabr: added, https://github.com/VocaDB/vocadb-loc/pull/2

VocaDbModel/Domain/ReleaseEvents/EventCategory.cs
```diff
@@ -24,5 +24,9 @@ public enum EventCategory
 		/// </summary>
 		Convention = 1 << 5,
 		Other = 1 << 6,
+		/// <summary>
+		/// J-POP Cover Festival
+		/// </summary>
+		Festival = 1 << 7,
```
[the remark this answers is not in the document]
kateabr: done

VocaDbWeb/wwwroot/locales/ja/VocaDb.Web.Resources.Domain.ReleaseEvents.json
```diff
@@ -5,6 +5,7 @@
     "Club": "クラブ",
     "Concert": "コンサート",
     "Contest": "コンテスト",
+	"Festival": "祭り",
```
ycanardeau, line 8: Which should we use, "祭" or "祭り"?

Tests/Domain/ReleaseEvents/ReleaseEventTypesTests.cs
```diff
@@ -0,0 +1,22 @@
...
+	[DataRow(8, EventCategory.Concert)]
+	[DataRow(16, EventCategory.Contest)]
+	[DataRow(32, EventCategory.Convention)]
+	[DataRow(64, EventCategory.Other)]
+	[DataRow(128, EventCategory.Festival)]
+	[TestMethod]
+	public void Value(int expected, EventCategory actual)
+	{
+		((int) actual).Should().Be(expected);
+	}
```
ycanardeau, line 21: Since the values of the `EventCategory` enum are saved as a string in [DB](https://github.com/VocaDB/vocadb/blob/d7e30980f28a03c95f06877e4d5c7075fc72ac80/VocaDb.Migrations/Migrations.cs#L717-L736), and not as a bit field, you should write tests against names instead of numbers (and can you please make sure that?). See [PVServiceTests.cs](https://github.com/VocaDB/vocadb/blob/94e53c2bbf075415a42cb4859e46bbbee3bd4229/Tests/Domain/PVs/PVServiceTests.cs).

VocaDbWeb/wwwroot/locales/ja/VocaDb.Web.Resources.Domain.ReleaseEvents.json
kateabr: ohh that's a typo, I wanted to write "祭"

Tests/Domain/ReleaseEvents/ReleaseEventTypesTests.cs
kateabr: Forgot to ask: if they are stored as strings, there should be no problem with switching the order order of values in the enum class, right? Currently, when selecting event type, "Festival" comes last, while I'd like it to come immediately after "Contest" (I think it makes more sense this way)

ycanardeau: Switching the order may work if and only if you are 100 percent sure that they are never stored as numbers. It would be disastrous if you accidentally changed it. I would suggest switching the order on the frontend instead, just like [ArtistTypesDropdownKnockout](https://github.com/VocaDB/vocadb/blob/9447fc42fed913e32b2c9f0d95eed103ce2a2545/VocaDbWeb/Scripts/Components/Shared/Partials/Artist/ArtistTypesDropdownKnockout.tsx), because it's more flexible. You could define a varaible like:
```tsx
const eventCategoryOrders: Record<EventCategory, number> = {
    [EventCategory.Unspecified]: 1,
    [EventCategory.AlbumRelease]: 2,
   // ...
}
```

ycanardeau:
Thanks for the PR!

PR #1223, opened 2022-09-17
Restore play queue items and index from the previous session

ycanardeau:


ycanardeau:
TODO: Prevent videos from automatically playing when opening in a new tab.

PR #1305, opened 2023-03-15
Replace the piapro embed player: change from swf to HTML5 iframe

szc126:
image

I have confirmed that the new .tsx file works, but I cannot tell where the .cshtml file is used.


szc126:
superseded by 9d9104452f26f527553eaa03d97c34d9aed57adb (?)

Issue #851, opened 2021-04-27
Other: Migrate to Bootstrap 5


FinnRG:
Split from #6.

Bootstrap 5 beta 3 was released on March 22, 2021 and the stable release is expected to release in the next few months (https://github.com/twbs/release and https://github.com/twbs/bootstrap/issues/31057). 

A migration to v5 would make it much easier to increase responsiveness for mobile devices and facilitate the implementation of new frontend features.

ycanardeau:
@Pyther99 Is there any reason you want to stick with Bootstrap? There are some other frameworks like [Material-UI](https://material-ui.com/) and [Primer](https://primer.style/). Think twice and make some experiments before you start working on this. Migrating to a new CSS framework is lower priority compared to migration to React anyway. Before creating a pull request, you can (and/or should) do mockups how the user interface would look, and post screenshots here or on Discord. I'll merge the PR if and only if the design looks better than the current one. (I'm quite conservative when it comes to CSS design.)

Pyther99:
@ycanardeau I would choose Bootstrap 5 as the aim of the migration is not to change the design of the site, this definitely requires intensive testing and consultation on Discord, but to make the site mobile friendly and keep the current design. As VocaDB currently uses Bootstrap 2, this can best be achieved by continuing to use Bootstrap, apart from that it is a well proven framework and I don't really see a good reason to choose another framework. After the React migration, https://github.com/react-bootstrap/react-bootstrap could also be used (which, however, is still based on Bootstrap 4.6).

PR #1312, opened 2023-04-02
Base for Mantine Conversion

FinnRG:
Start of #851. @ycanardeau and I decided that it would be best to do the frontend migration with Mantine instead of Bootstrap 5. This PR is just an example how one incremental step in the migration process might look like and to gather feedback on how this process should be structured. 

One unsolved problem is if and how an UI switch should be implemented. Maybe we could use the existing theme mechanics to optionally load the bootstrap css, but I don't have enough insight to determine if this is a viable solution.
Or maybe the best solution is to just do the migration in a separate branch and to merge once the migration is finished.

Screenshot from 2023-04-02 18-46-46
(Obviously this still needs some improvements like colors and a better layout, but it shows that a conversion to Mantine is feasible) 

The benefits of a conversion would be:

1. Improved developer experience
2. Improved accessibility
3. Improved responsiveness
4. (Improved performance)

### DIFF
--- a/VocaDbWeb/Scripts/App.tsx
+++ b/VocaDbWeb/Scripts/App.tsx
@@ -11,6 +11,7 @@ import { LoginManagerProvider } from '@/LoginManagerContext';
 import { MutedUsersProvider } from '@/MutedUsersContext';
 import { VdbProvider } from '@/VdbContext';
 import '@/i18n';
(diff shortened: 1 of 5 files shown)


FinnRG:
Closed in favor of #1399.

PR #1313, opened 2023-04-11
bilibili, fix unhandled exception for unviewable videos

szc126:
I am not familiar with C# so this code needs a good review first.

https://vocadb.net/api/pvs and https://vocadb.net/api/songs/findDuplicate would return "Sorry, an unhandled error occurred :(" instead of expected JSON data.

The log was like this:

```
[.NET ThreadPool Worker] INFO Loading Bilibili URL http://bilibili.com/video/av575832 
[.NET ThreadPool Worker] ERROR An unhandled exception has occurred while executing the request. - System.NullReferenceException Object reference not set to an instance of an object. Void MoveNext()    at VocaDb.Model.Service.VideoServices.VideoServiceBilibili.ParseByUrlAsync(String url, Boolean getTitle) in /tmp/vocadb-dev/git/vocadb/VocaDbModel/Service/VideoServices/VideoServiceBilibili.cs:line 53
   at VocaDb.Model.Database.Queries.SongQueries.FindDuplicates(String[] anyName, String[] anyPv, Int32[] artistIds, Boolean getPVInfo) in /tmp/vocadb-dev/git/vocadb/VocaDbModel/Database/Queries/SongQueries.cs:line 946
   at VocaDb.Web.Controllers.Api.SongApiController.GetFindDuplicate(String[] term, String[] pv, Int32[] artistIds, Boolean getPVInfo) in /tmp/vocadb-dev/git/vocadb/VocaDbWeb/Controllers/Api/SongApiController.cs:line 127
   at lambda_method1024(Closure, Object)

[...]
```

### DIFF
--- a/VocaDbModel/Service/VideoServices/VideoServiceBilibili.cs
+++ b/VocaDbModel/Service/VideoServices/VideoServiceBilibili.cs
@@ -50,9 +50,9 @@ public class VideoServiceBilibili : VideoService
 			return VideoUrlParseResult.CreateError(url, VideoUrlParseResultType.LoadError, new VideoParseException($"Unable to load Bilibili URL: {x.Message}", x));
 		}
 
-		// cannot view video
+		// If the video cannot be viewed,
 		// web returns the error message "啊叻？视频不见了？" (あれ? the video's gone?)
-		// api simultaneously returns the message "稿件不可见" (upload cannot be viewed)
+		// API simultaneously returns the message "稿件不可见" (upload cannot be viewed)
 		if (response.Code == 62002)
 		{
 			return VideoUrlParseResult.CreateError(url, VideoUrlParseResultType.LoadError, "Video cannot be viewed");
--- a/VocaDbModel/Service/VideoServices/VideoServiceBilibili.cs
+++ b/VocaDbModel/Service/VideoServices/VideoServiceBilibili.cs
@@ -53,7 +53,7 @@ public class VideoServiceBilibili : VideoService
 		// If the video cannot be viewed,
 		// web returns the error message "啊叻？视频不见了？" (あれ? the video's gone?)
 		// API simultaneously returns the message "稿件不可见" (upload cannot be viewed)
-		if (response.Code == 62002)
+		if (response == null || response.Data == null || response.Code == 62002)
 		{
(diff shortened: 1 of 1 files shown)


VocaDbModel/Service/VideoServices/VideoServiceBilibili.cs
```diff
@@ -105,6 +113,7 @@ public class BiliMetadata
 
 class BilibiliResponse
 {
+	public int Code { get; init; } = default!;
 	public BilibiliResponseData Data { get; init; } = default!;
```
FinnRG, line 117: The `default!` expression means, that Code and Data can't be null, but in your error case Data should be null (because it isn't present in the json response). Could you change Data to be nullable and check if this causes errors somewhere else in the code?

FinnRG (changes requested):
Thank you for your PR. I've only found one small issue.

VocaDbModel/Service/VideoServices/VideoServiceBilibili.cs
szc126: Thank you for pointing that out! With zero C# experience, I am frankly not very sure how to do that. From what I can tell, it would be `BilibiliResponseData? Data { get; init; } = default;`? Is that right?

FinnRG: Yes, that is correct

VocaDbModel/Service/VideoServices/VideoServiceBilibili.cs
```diff
@@ -50,6 +50,14 @@ public override async Task<VideoUrlParseResult> ParseByUrlAsync(string url, bool
 			return VideoUrlParseResult.CreateError(url, VideoUrlParseResultType.LoadError, new VideoParseException($"Unable to load Bilibili URL: {x.Message}", x));
 		}
 
+		// cannot view video
+		// web returns the error message "啊叻？视频不见了？" (あれ? the video's gone?)
+		// api simultaneously returns the message "稿件不可见" (upload cannot be viewed)
+		if (response.Code == 62002)
```
FinnRG, line 56: This check should also be changed to `if (response == null || response.Data == null || response.Code == 62002)`, because now both the response object and the `response.Data` object could be null.

FinnRG:
Thank you for your contribution.

PR #1431, opened 2023-05-18
Improve alignment of buttons, icons and text in navigation bar buttons

Dyn39:
Before:
image
image
image

After:
image
image
image


FinnRG:
Thank you for your contribution!

Issue #1585, opened 2023-10-04
Songs: Demo/pre-release song type as a base version of "Original" songs


Susko3:
### Is your feature request related to a problem? Please describe. (required)
Some songs in the database have `Original version` relationships that don't make chronological sense or don't follow producer intent. These are two separate problems that are caused by `Original song`s being the root of the derivatives graph.

#### Examples:

Here's how the current derivatives graph looks like for [Hello, Again](https://vocadb.net/S/201986):

```mermaid
flowchart LR
 O{{"Hello, Again
 Original song"}}
 OI("Hello, Again
 Instrumental")
 D("Hello, Again (Demo Version)
 Remaster")
 DI("Hello, Again (Demo Version)
 Instrumental")
 C("Hello, Again (Oliver cover)
 Cover")

 O -->D & C & OI
 D --> DI
```

I propose that the following graph would be more appropriate as it makes more chronological sense. Since the demo version has definitely existed before the full release version.

```mermaid
flowchart LR
 O{{"Hello, Again
 Original song"}}
 OI("Hello, Again
 Instrumental")
 D("Hello, Again (Demo Version)
 Demo")
 DI("Hello, Again (Demo Version)
 Instrumental")
 C("Hello, Again (Oliver cover)
 Cover")

 D --> DI & O
 O --> C & OI
```

---

The second problem is when VocaDB doesn't follow artist intent. Take the graph of [Puzzle by CircusP](https://vocadb.net/S/380012):

```mermaid
flowchart LR
 O("Puzzle feat. Miku
 YELLOW
 Cover")
 P{{"Puzzle feat. GUMI
 Puzzle (Patreon Release)
 Original song"}}
 C("Puzzle (Yin-P cover)
 Cover")
 PC("Puzzle feat. Aido
 Puzzle (Patreon Release)
 Cover")
 I("Puzzle
 Puzzle (Patreon Release)
 Instrumental")

 P --> O & PC & I
 O --> C
```

Nobody would consider the public release of Puzzle to be a cover song. CircusP said in his streams that he considers the public release to be the main version and should therefore be the `Original song`:

```mermaid
flowchart LR
 O{{"Puzzle feat. Miku
 YELLOW
 Original song"}}
 P("Puzzle feat. GUMI
 Puzzle (Patreon Release)
 Pre-release")
 C("Puzzle (Yin-P cover)
 Cover")
 PC("Puzzle feat. Aido
 Puzzle (Patreon Release)
 Cover")
[...]

PR #1587, opened 2023-10-05
Allow original songs to be derived from another song

Susko3:
- First part of the solution to #1585

As it turns out, original/base versions of original songs can already be added to the database, just that there's a bunch of special casing to try to hide that fact from the frontend and the API.

This PR just removes the special cases, allowing:
- original song pages to show their original/base versions under "Original version" (including API)
- base song pages to show the original song under "Alternate versions"

**Things that are currently missing from this PR:**
- UX considerations for the text shown, as having an original version of an original song is a bit cumbersome

# Breaking changes

## Original songs that have an original version entered in the database will show that in the UI

People could accidentally enter in an original version (and not realise it, as it's hidden). Those entries should be removed/fixed before this PR makes it in.

I'd like to see which songs are affected. In the spirit of not DoS-ing VocaDB, this is best queried directly from the database:
```sql
SELECT Id, EnglishName, ArtistString, OriginalVersion FROM Songs
WHERE SongType = 'Original' AND OriginalVersion IS NOT NULL;
GO
```

# Behaviour on current version of live website

You can already add an original version to an original song on the _live website_:

1. On the Edit page, change the song type to not be Original song
2. Select the desired song to be the Original version
3. Change song type back to Original song (while the GUI is invisible, the original version remains selected)
4. Save changes

You can see a song changed this way in https://vocadb.net/Song/ViewVersion/1714212.
But it won't show on the [main song page](https://vocadb.net/S/201986).

image

Note the absence of `originalVersionId` in the [API request for the original song](https://vocadb.net/api/songs/201986)

image

and the inclusion of the original song in the [base's derived songs](https://vocadb.net/api/songs/201987/derived).

image

### DIFF
--- a/VocaDbWeb/Scripts/Stores/Song/SongDetailsStore.ts
+++ b/VocaDbWeb/Scripts/Stores/Song/SongDetailsStore.ts
@@ -308,7 +308,7 @@ export class SongDetailsStore {
 
 		this.tagUsages = new TagListStore(data.tagUsages);
 
-		if (data.songType !== SongType.Original && !this.originalVersion.entry) {
+		if (!this.originalVersion.entry) {
 			this.getOriginal(data.linkedPages);
 		}
 	}
--- a/VocaDbModel/Domain/Songs/Song.cs
+++ b/VocaDbModel/Domain/Songs/Song.cs
@@ -227,7 +227,7 @@ public class Song :
 	/// This method also tests that the song type isn't "Original", because original songs aren't supposed to have parent songs.
 	/// </summary>
 	[MemberNotNullWhen(true, nameof(OriginalVersion))]
-	public virtual bool HasOriginalVersion => SongType != SongType.Original && OriginalVersion != null;
+	public virtual bool HasOriginalVersion => OriginalVersion != null;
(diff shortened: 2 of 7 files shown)


VocaDbWeb/Scripts/Stores/Song/SongDetailsStore.ts
```diff
@@ -308,7 +308,7 @@ export class SongDetailsStore {
 
 		this.tagUsages = new TagListStore(data.tagUsages);
 
-		if (data.songType !== SongType.Original && !this.originalVersion.entry) {
+		if (!this.originalVersion.entry) {
 			this.getOriginal(data.linkedPages);
 		}
```
Susko3, line 313: This will cause original songs that have unofficial links to another original song on VocaDB (or UtaiteDB/TouhouDB) to have the latter song considered the original version (this would only be shown in the UI and wouldn't be reflected in the database).

Example: https://vocadb.net/S/201986 (this one actually doesn't matter in the end as the song will have a properly defined demo version).

Could also be checked with an SQL query to the database.

PR #1580, opened 2023-08-28
refactor: Use th new Nextjs app router

FinnRG:
This should greatly improve the new frontend performance.

PR #1592, opened 2023-10-08
refactor: Migrate the new frontend to shadcn / tailwindcss

FinnRG:
After implementing #1580, it turns out that Mantine does not work well with NextJS's new App Router. Therefore, TailwindCSS with components from Shadcn is now used instead. This should improve performance and increase developer productivity.

Furthermore, the new theme system in Mantine creates unnecessary rerenders that led to flashing content and poor performance. With [Shadcn Themes](https://ui.shadcn.com/themes) these are prevented.


FinnRG:
This PR is now outdated. All of the necessary changes are included in #1754.

PR #1628, opened 2023-11-19
feat: New admin page with a list of deleted tags

FinnRG:
Trusted+ users commonly run into issues, where tag names have already been used for deleted tags. This new page will list every deleted tag.

Functionality of the `/Album/Deleted` page has also been restored.


jukkahyv:
Maybe you were already aware of it, but there was a page for listing deleted albums. It should be [here](https://touhoudb.com/Album/Deleted). However, it seems to be broken, at least on TouhouDB. I was just thinking you would probably want to make use of the same code, maybe even the same view for all deleted entries?

PR #1830, opened 2024-11-05
feat: /Details rate limiting

### LINKED ISSUE #1596: Rate limit to combat spam
### Is your feature request related to a problem? Please describe.
Recent events showed that some kind of rate limiting is a good solution to combat spam.
Multiple entries were created in a short time which should be prevented.

### Describe the solution you'd like.
As described here: https://discord.com/channels/309072240639737866/797103264655081492/1166197842726101063

Solutions could be:
- newly registered users can only create entries after some time passed
- a threshold on how many they can create in a specific amount of time
- email must be verified first before adding/editing entries

It was decided that a rate limit of **100 entries/h** makes the most sense.
This rule should only be for regular users/non-trusted users.

___

Related issues: https://github.com/VocaDB/vocadb/issues/1149 & https://github.com/VocaDB/vocadb/issues/32

FinnRG:
Closes #1596 and adds a 40 req/minute sliding window rate limit and a max. 4 concurrent requests limit for /details endpoints.

### DIFF
--- a/VocaDbWeb/Program.cs
+++ b/VocaDbWeb/Program.cs
@@ -1,3 +1,4 @@
+using Microsoft.AspNetCore.RateLimiting;
 using Autofac;
 using Autofac.Extensions.DependencyInjection;
 using NLog.Web;
@@ -17,6 +18,17 @@ try
 
 	startup.ConfigureServices(builder.Services);
 
+	builder.Services.AddRateLimiter(options =>
+	{
+		options.RejectionStatusCode = StatusCodes.Status429TooManyRequests;
+		options
+			.AddSlidingWindowLimiter(policyName: "details", options =>
+			{
+				options.Window = TimeSpan.FromMinutes(1);
+				options.PermitLimit = 40;
+				options.SegmentsPerWindow = 1;
+			});
+	});
 	builder.Host.UseServiceProviderFactory(new AutofacServiceProviderFactory());
 	builder.Host.ConfigureContainer<ContainerBuilder>(startup.ConfigureContainer);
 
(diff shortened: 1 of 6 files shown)


VocaDbWeb/Program.cs
```diff
@@ -17,6 +18,17 @@
 
 	startup.ConfigureServices(builder.Services);
 
+	builder.Services.AddRateLimiter(options =>
+	{
+		options.RejectionStatusCode = StatusCodes.Status429TooManyRequests;
+		options
+			.AddSlidingWindowLimiter(policyName: "details", options =>
```
Susko3, line 25: `"details"` should probably be a `public const string` in a constants class instead of a magic value that needs to be specified in multiple places.

VocaDbWeb/Program.cs
```diff
@@ -17,6 +18,17 @@
...
 
+	builder.Services.AddRateLimiter(options =>
+	{
+		options.RejectionStatusCode = StatusCodes.Status429TooManyRequests;
+		options
+			.AddSlidingWindowLimiter(policyName: "details", options =>
+			{
+				options.Window = TimeSpan.FromMinutes(1);
+				options.PermitLimit = 40;
+				options.SegmentsPerWindow = 1;
```
Susko3, line 29: This code doesn't match the PR description.

> 40 req/sec 

```cs
TimeSpan.FromMinutes(1)
```

> max. 4 concurrent requests 

Nowhere to be seen.

Susko3:
How does the frontend deal with / show the user the HTTP status 429 Too Many Requests?

VocaDbWeb/Program.cs
FinnRG: That's why it's still a draft PR. (You've probably also noticed that the rate limit is currently global and partitioned by the IP address of the user agent)

FinnRG:
> How does the frontend deal with / show the user the HTTP status 429 Too Many Requests?

There will be a custom 429 error using the same mechanism as the other error pages.

Issue #1378, opened 2023-04-26
Multiple entries: Sometimes, previews (tooltips) of artists do not display correctly


andreoda:
### Website
VocaDB

### Describe the bug
Even after clearing caches and such, the site will sometimes not load the correct previews (tooltips) of artists when for example inspecting artists from album songs.

I couldn't figure out a pattern yet to reproduce this.

### Expected behavior
Previews showing the correct information.

### Screenshots
Wrong display:
![grafik](https://user-images.githubusercontent.com/52045122/234639437-365631e5-26bb-47ff-9950-98268c1784b2.png)

Correct display:
![grafik](https://user-images.githubusercontent.com/52045122/234639528-0d0c65f9-8f71-47ed-b0fa-7d82f47f9642.png)

### Desktop (please complete the following information):
 - OS: Windows
 - Browser: Firefox

lalondeguy:
I've noticed this also happens often when looking at alternate versions of a song? Such as here (pic taken on Masa Works Design's remix of Alien Alien):
![image](https://github.com/VocaDB/vocadb/assets/109246223/a5df8fab-42f1-43da-977b-73ad04974c17)

lalondeguy:
Again on different entries (Pepoyo's fan MV of Obake no Shoutaijou and the Project Sekai cover of Dance Orchestra), I believe this only happens when the role of the highlighted individual matches the role of a previously highlighted one, since I tried on another song (Salmon-Vanilla's fan MV of Your Adventure Log Has Vanished's Sekai cover after highlighting both Jizasu-P and Glider in separate instances, as well as highlighting Tsukasa Tenma on the official MV of said Sekai cover after highlighting Jizasu-P on the original version) and could not replicate it.
![image](https://github.com/VocaDB/vocadb/assets/109246223/7f28b943-2cd4-45bc-a761-ff387d9329ee)
![image](https://github.com/VocaDB/vocadb/assets/109246223/ebe6e51d-7531-4cd3-ba70-25ddb171bb3a)

lalondeguy:
This also exclusively seems to happen in songs that are directly linked and do not refresh the page when clicked, as I was unable to replicate this by highlighting a producer and then highlighting a separate producer on a song unaffiliated with the previous one, and I was also unable to replicate it by clicking a song that was linked in another song's description, which refreshes the page.

Shiroizu:
To reproduce:

[...]

PR #1871, opened 2024-12-24
Bugfix 1378 Tooltip incorrect display.

Calamitish:
Tooltip currently does not update value (e.g. artist) is currently already in state. This can effectively be solved by no longer immediately returning when the value is known, but also comparing its id against the current supplied id.

Closes https://github.com/VocaDB/vocadb/issues/1378

### DIFF
--- a/VocaDbWeb/Scripts/Components/KnockoutExtensions/EntryToolTip.tsx
+++ b/VocaDbWeb/Scripts/Components/KnockoutExtensions/EntryToolTip.tsx
@@ -45,7 +45,7 @@ export const AlbumToolTip = React.memo(
 		const [album, setAlbum] = React.useState<AlbumContract>();
 
 		React.useEffect(() => {
-			if (album) return;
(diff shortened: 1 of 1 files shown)


Susko3:
It's useful to mention the issue (#1378) in the PR description so that GitHub can link the two together and make it easier to navigate to the relevant issue.

https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue

Calamitish:
> It's useful to mention the issue (#1378) in the PR description so that GitHub can link the two together and make it easier to navigate to the relevant issue.
> 
> https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue

Got it!
I mainly use bitbucket/gitlab so I'm real rusty on github, haha.

FinnRG:
@Calamitish Thank you for the fix!

PR #1873, opened 2024-12-30
Change langversion to latest instead of preview for VocaDb.Model

riipah:
I noticed the build was failing in AppVeyor and also my machine. VocaDb.Model project was set to langversion = preview, which enables "field" as reserved keyword. I think preview should only be used temporarily, because it can be unstable. It was probably set years ago to enable features that were preview at that time, and it's no longer needed. Microsoft recommends setting the language version to fixed value (such as 13), or removing it to use default, but I didn't want to start experimenting at this point. langversion = latest is better than preview, and it fixes the build, on my machine at least.


FinnRG:
@riipah Thank you. I'll use this branch to fix the failing test (from the tag target changes) and then merge it.

Issue #1886, opened 2025-01-25
Account / User profile / Front page: Add the "Artist Verification Page" to the account dropdown menu and/or the sidebar


Bananapafo:
### Is your feature request related to a problem? Please describe. (required)

The "Artist Verification Page" is only available through the user's account profile (which disappears after verifying) and [this vocadb editing wiki page](https://wiki.vocadb.net/docs/artists/artist-verification) (Which I'm not sure if a lot of people even check the wiki). There's a lot of accounts made by producers that haven't verified their accounts to be linked with their producer entries (this includes voicebanks).

### Describe the solution you'd like. (required)

I think adding the link to the account dropdown menu and/or the sidebar would encourage more users to verify their accounts to their related entries. It would also make it easier for them to verify their accounts to voicebanks they made after they verified their accounts initially. The link should primarily be added to the account dropdown menu, since the sidebar isn't visible on the mobile version of the side, but I think adding them both would be good.

### Describe alternatives you've considered. (optional)

Haven't found any alternatives so far.

### How should the UI look like? (illustrate it) (optional)

Here's how it could look in the dropdown menu

![Image](https://github.com/user-attachments/assets/024ba0b0-e945-4499-a2ed-34d66e3622ac)

And in the sidebar (Made by HUK when I suggested the idea in the discord)

![Image](https://github.com/user-attachments/assets/919990c5-cfba-4355-81a0-ccbe410d674b)

### Checklist (required)

Fill out the checklist.

- [x] Searched **[VocaDB/vocadb](https://github.com/VocaDB/vocadb/issues)** for duplicate issues.

mn7216:
I don't see much point in adding it to the user dropdown. It should be fine just being in the sidebar.
sidebar addressed in [this pull request](https://github.com/VocaDB/vocadb/pull/1887)

PR #1887, opened 2025-01-25
Add artist verification to sidebar

mn7216:
adds artist verification link to the sidebar as per #1886, tested on beta site and works


.gitignore
```diff
@@ -366,4 +366,6 @@ FodyWeavers.xsd
 
 db/
 docker-compose.yml
-wwwroot
\ No newline at end of file
+wwwroot
+song_rating_scores_2024-12-31.csv
+VocaDbWeb/vite.config.ts
```
FinnRG, line 371: Could you remove these two entries? We want to track `vite.config.ts` and the CSV seems specific to your local setup

FinnRG (changes requested):

.gitignore
mn7216: > Could you remove these two entries? We want to track `vite.config.ts` and the CSV seems specific to your local setup

Done

PR #1896, opened 2025-02-02
chores: Broaden the NND MyList link matcher

szc126:
Match both
- https://www.nicovideo.jp/mylist/000 (old URL style) and
- https://www.nicovideo.jp/user/000/mylist/000 (new URL style)

### DIFF
--- a/VocaDbWeb/Scripts/Shared/WebLinkMatcher.ts
+++ b/VocaDbWeb/Scripts/Shared/WebLinkMatcher.ts
@@ -944,7 +944,10 @@ export class WebLinkMatcher {
 			cat: WebLinkCategory.Official,
 		},
 		{
-			url: 'nicovideo.jp/mylist/',
+			// matching both
+			// https://www.nicovideo.jp/mylist/000 (old URL style) and
+   			// https://www.nicovideo.jp/user/000/mylist/000 (new URL style)
+			url: '/mylist/',
 			desc: 'NND MyList',
 			cat: WebLinkCategory.Official,
 		},


VocaDbWeb/Scripts/Shared/WebLinkMatcher.ts
```diff
@@ -944,7 +944,10 @@ export class WebLinkMatcher {
 			cat: WebLinkCategory.Official,
 		},
 		{
-			url: 'nicovideo.jp/mylist/',
+			// matching both
+			// https://www.nicovideo.jp/mylist/000 (old URL style) and
+   			// https://www.nicovideo.jp/user/000/mylist/000 (new URL style)
+			url: '/mylist/',
```
FinnRG, line 950: I would prefer to have two entries in `WebLinkMatcher.ts` for the new and the old style rather than including this one matcher without a host

FinnRG (changes requested):

VocaDbWeb/Scripts/Shared/WebLinkMatcher.ts
szc126: I initially did that, but then I realized that the one with the host would be redundant as the short one would match a superset; but I also understand the argument that it's stylistically better

FinnRG: I'm really sorry for the late reply. My main argument for two separate entries is to still keep the host (https://www.nicovideo.jp) in the matcher to keep URLs who just happen to include `/mylist/` from matching with this matcher

szc126: My understanding is that
- `https://www.nicovideo.jp/mylist/000` can be matched with `nicovideo.jp/mylist/` (current behavior), but
- `https://www.nicovideo.jp/user/000/mylist/000` must be matched with the host-less `/mylist/` (proposed edit), and cannot have a host (ideally `nicovideo\.jp/user/\d+/mylist/` if represented in regex) due to the simplistic use of `indexOf` (not regex) that cannot handle the necessary wildcard for the intervening `/user/000/`, and
- `nicovideo.jp/mylist/` would then be a redundant test to `/mylist/`.

This \*does\* allow "false positives" for other hosts, but I am not sure if this would happen in practice.

PR #1901, opened 2025-03-05
Add Amdo Tibetan

mn7216:


szc126:
We have so few Tibetan content that I wonder if this is necessary or user-friendly. We already merged Quechua/Kichwa. I think the type of Tibetan could be simply specified in the description.

mn7216:
> We have so little Tibetan content that I wonder if this is necessary or user-friendly. We already merged Quechua/Kichwa. I think the type of Tibetan could be simply specified in the description.

I've added it because I received a request asking about it. It seems odd to reject one in particular when there are many relatively obscure dialects/languages already here.

PR #1924, opened 2025-04-24
Refactor WebLinkMatcher.ts and add additional.

Calamitish:
# ExtLinks.css:
- Replace all double quotes with single quotes for consistency.
- Add trailing slashes to all URLs for consistency.
- Clean up usage of background-image vs background.
- Unify http and https checks. Before this, some icons would only match with http URLs and some only with https URLs.
- Make bracket usage consistent.
- Remove or combine duplicate entries.
- Removed a dead comment.
- Re-order ExtLinks.css by logo image name alphabetically.
- - With exception for those those that have a logo in a different repository, move those to the top.
- Rename fasic and thbwiki to Fasic and THBWiki to match actual filenames. naming convention for files could be a discussion point for a future PR.
- Add icon matching with links that did have icons, but were not being used.
- Add additional icon matching with new icons (see: https://github.com/VocaDB/ExtIcons/pull/7)
- Icon matching via ^ match or * match has been left mostly the same, could still be a discussion point for a future PR.

# WebLinkMatcher.ts
- Add additonal Weblink matchers relating to https://github.com/VocaDB/ExtIcons/pull/7


Susko3 (changes requested):
To me, this PR is impossible to review. It's possible that you accidentally changed a link category or dropped some icon. It would be way easier to check that if this PR was split into separate commits that do one of the steps you mention in the description.

Also, every line in the diff is marked as changed, even if it remained the same. Have the line-endings accidentally been changed? Maybe make the line-ending change a separate commit, or don't do it. https://docs.github.com/en/get-started/git-basics/configuring-git-to-handle-line-endings

Calamitish:
> every line in the diff is marked as changed
Seems like the project is operating with LF line endings which automatically got converted to CRLF on my end.
Will revert and separate in different commits.
That last one's still going to be a pain to review just by nature of sorting all of the ExtLinks alphabetically.

Calamitish:
PR updated and everything is now split into separate commits.
Let me know if I've missed something.

Something to note:
It seems that the project either inconsistently uses LF and CRLF line-endings or only uses LF.
(It's very hard for me to tell, in all honesty)
It's probably a good idea to either outline this in the wiki entry for [windows development environment](https://wiki.vocadb.net/docs/development/development-environment-windows) or the [project readme](https://github.com/VocaDB/vocadb/blob/main/README.md).
It's currently easy to miss this difference for any contributors.
Might also be a good idea to sweep the entire project and double check that all files consistently use the same line-endings.

VocaDbWeb/Scripts/styles/ExtLinks.css
```diff
@@ -1,900 +1,938 @@
 a.extLink {
```
Susko3, line 1: Inconsistent brace style.

Susko3 (changes requested):
I've checked that the individual commits make changer per their description. Everything else seems fine, apart from the following:

VocaDbWeb/Scripts/styles/ExtLinks.css
```diff
@@ -1,900 +1,938 @@
...
+	background-image: url('/Content/ExtIcons/touhouwiki.png');
 }
 
-a.extLink[href*='tinami.com'],
-a.extLink[href*='tinami.jp'] {
-	background-image: url('/Content/ExtIcons/tinami.png');
+a.extLink[href^='http(s?)://tower.com/'],
+a.extLink[href^='http(s?)://tower.jp/']
+{
+	background-image: url('/Content/ExtIcons/towerrecords.png');
```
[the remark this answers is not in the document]
Calamitish: Seems I eye'd over the www. a bit too quickly there, good catch, on it!

VocaDbWeb/Scripts/styles/ExtLinks.css
```diff
@@ -1,900 +1,938 @@
...
 
-a.extLink[href^="http://about.me/"],
-a.extLink[href^="https://about.me/"]
+a.extLink[href^='mailto:/']
+{
+	background-image: url('/Content/Icons/email.png');
+}
+
+/* image outside of ExtIcons folder because it is also used elsewhere*/
+a.extLink[href^='http(s?)://soundcloud.com/'],
```
Susko3, line 16: Have you tested this? Regex doesn't seem to be supported in CSS [attribute selectors](https://developer.mozilla.org/en-US/docs/Web/CSS/Attribute_selectors).

slika

I suggest you either
a) go back to separate `http://` and `https://`
b) change the matcher to be `[href*='://soundcloud.com/']` (anywhere in url, but including `://` to avoid false-positives)

Calamitish: Went with b as to not have unnecessary duplication.

VocaDbWeb/Scripts/styles/ExtLinks.css
```diff
@@ -1,900 +1,940 @@
...
 }
 
-a.extLink[href*='www.facebook.com/'] {
-	background: url('/Content/ExtIcons/facebook.png') no-repeat left 50%;
+a.extLink[href*='://www.metal-archives.com/']
+{
+	background-image: url('/Content/ExtIcons/ma.png');
 }
 
-a.extLink[href^="https://plus.google.com/"]
```
Susko3, line 460: still regex and kinda very wrong

VocaDbWeb/Scripts/styles/ExtLinks.css
```diff
@@ -1,900 +1,940 @@
...
-a.extLink[href^="http://com.nicovideo.jp/community/"],
-a.extLink[href^="https://com.nicovideo.jp/community/"],
-a.extLink[href*='ch.nicovideo.jp'] {
-	background-image: url('/Content/nico.png');
+a.extLink[href*='://musicbrainz.org/']
+{
+	background-image: url('/Content/ExtIcons/musicbrainz.png');
 }
 
-a.extLink[href^="http://www.nicovideo.jp/"],
```
Susko3, line 522: ditto

Susko3 (changes requested):

Susko3:
Whoops, the review went wrongly, just search for "?" and you'll see the problems.

Calamitish:
Got them!
Here's to writing code when you're tired :D
I probably typo'd those originally so they didn't get caught be the replace.

VocaDbWeb/Scripts/styles/ExtLinks.css
```diff
@@ -1,900 +1,940 @@
...
 
-a.extLink[href^="http://www5.atwiki.jp/hmiku/"],
-a.extLink[href^="https://www5.atwiki.jp/hmiku/"],
-a.extLink[href^="https://w.atwiki.jp/hmiku/"],
-a.extLink[href*="atwiki.jp/cevio_synthv/"] {
-	background-image: url('/Content/ExtIcons/atwiki.png');
+a.extLink[href*='://web.archive.org/']
+{
+	background-image: url('/Content/ExtIcons//archive.png');
 }
```
Susko3, line 122: I think this should be the last rule in the file, so that it takes precedence over all other rules. Currently, there are inconsistencies:

slika

Calamitish: Surprised that this always seemed to work fine before.
It used to hang around line 70 or so.

Susko3 (approved):
Seems to be good now!

FinnRG (approved):
Thank you both for your work on this.

PR #1957, opened 2025-05-13
chore: Add Sakha

szc126:
https://vocadb.net/Ar/161363


Susko3 (approved):
Matches the ISO 639-2 and ISO 639-3 codes: https://iso639-3.sil.org/code/sah. There isn't an [ISO 639-1 code](https://www.loc.gov/standards/iso639-2/php/langcodes_name.php?code_ID=382) for Yakut.

Issue #437, opened 2018-10-27
Front page: Proper credit link for background illustration based on style


riipah:
Since different styles have different backgrounds, the link should also change. Currently hardcoded to default style link.

PR #2028, opened 2026-01-14
feat: New UtaiteDB background image

FinnRG:


FinnRG:
This also implements the necessary code to fix #437, but I still need to find the proper artist credits for the background images.

Issue #1386, opened 2021-07-08
Songs: Add field to copy-paste the video description


andreoda:
For song entries

**Is your feature request related to a problem? Please describe.**
- people forget to add information
- people can't find information (they don't look into the video description)
- video got deleted and info/video description is missing now

**Describe the solution you'd like**
A field meant to put the video's description in by copy-paste. From video sites like YouTube, NicoNicoDouga and bilibili.

**Describe alternatives you've considered**
The normal description field is not there to copy-paste the video description, since it is (in most times) written in first person.
The "artist's personal description" field is only meant for registered and verified artists/producer to put their own (first person) description in.

Maybe expand the usage for the "artist's personal description", so that it is allowed to copy-paste the video description but that is not possible since only verified artists and mods are allowed to use this.

**Additional context**
If a new field is created for this, maybe the system could automatically input the video description. If a video has two or three uploads or are on different sites (YT and NND) that would not work though.

PR #2043, opened 2026-01-29
feat: Automatic PV description metadata for YouTube, Bilibili and Vimeo

FinnRG:
Part of #1386. This adds automatic description fetching for YouTube, Bilibili and Vimeo.

NND and Bandcamp will be added soon, but I'll first need to add their API client repos to the source tree.

### DIFF
--- a/VocaDbModel/DataContracts/PVs/PVContract.cs
+++ b/VocaDbModel/DataContracts/PVs/PVContract.cs
@@ -97,8 +97,8 @@ public class PVContract : IPVWithThumbnail
 
 	/// <summary>
 	/// Video description from the source service.
-	/// Not serialized in API responses.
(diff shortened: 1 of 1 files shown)


szc126:
Awesome. But something I wonder upon looking at this is: what if we captured description \*and\* view count, \*periodically\* (after every <var>n</var> page visits, or something), like our own Vocaloid-only nicolog clone? I think it would be even more powerful (recording view count could remove the need for the view count pools, for example). (Please forgive me for yapping without contributing to implementation...)

FinnRG:
@szc126 I definitely want to capture view counts in the future, but to make that really useful we should first add some kind of PV metadata history service that tracks changes and periodically refreshes data.

Issue #1970, opened 2025-07-23
Albums: Show 'no rating' instead of empty stars in album review tab


FinnRG:
### Is your feature request related to a problem? Please describe. (required)
The current UI shows 0 stars if a user reviewd an album without assigning a score. This makes it look like the user has given the album a 0 star rating.

<img width="1549" height="215" alt="Image" src="https://github.com/user-attachments/assets/961481a0-95c9-4a4e-ba78-76651c7bdcfa" />

### Describe the solution you'd like. (required)
The UI should instead don't show the stars at all and maybe display an info label.

### Describe alternatives you've considered. (optional)
We could also force users to give a rating score in their review, but I would rather give the users the freedom to express their rating as text without assigning a score.

### Checklist (required)

Fill out the checklist.

- [x] Searched **[VocaDB/vocadb](https://github.com/VocaDB/vocadb/issues)** for duplicate issues.

PR #2144, opened 2026-04-06
Default text if no star rating is given

C0unterPo1nt:
#1970 Simple feature add. 

image

First time contribution. Feel free to roast me if something needs fixing. I used "no rating given" instead of "no rating" because I felt the text was clearer.


FinnRG (approved):
Thank you for your contribution, the changes all look good 👍

FinnRG:
Minor note, but I would recommend using a standard like conventional commits and squashing fixup commits. I will do a squash merge anyway so it's not a requirements, just a recommendation.

FinnRG:
I would also recommend writing "Fixed #id" in the MP, so that the bug is automatically closed when merging the MP.

Issue #2056, opened 2026-02-06
Albums: Half star ratings


FinnRG:
### Is your feature request related to a problem? Please describe. (required)
We currently only allow 0-5 stars for album ratings. Some users would like to create more detailed reviews.

### Describe the solution you'd like. (required)
We should allow half star ratings for album reviews.

### Describe alternatives you've considered. (optional)
We could also go to a number / percentage based system: 80/100

### Checklist (required)

Fill out the checklist.

- [x] Searched **[VocaDB/vocadb](https://github.com/VocaDB/vocadb/issues)** for duplicate issues.

PR #2167, opened 2026-05-20
Issue 2056: Half-star Ratings for Albums

tonyhieu:
This PR addresses [Issue 2056](https://github.com/VocaDB/vocadb/issues/2056) by allowing users to submit half star ratings. In addition, ratings on pages are fractional as opposed to being whole stars. Below are screenshots taken from a local build of the PR:

Albums list under artist:
image

Album page:
image

Update pop-up:
image

This PR includes both front-end and back-end changes. The front-end changes are seen above. Notably, the `EditCollectionDialog` component for inputting a rating has a new import: the outdated `import JqxRating from 'jqwidgets-scripts/jqwidgets-react-tsx/jqxrating';` is replaced with a more modern `import Rating from '@mui/material/Rating';` from `material-ui`. For displaying fractional stars, the `<Stars />` component has been tweaked and is now used in places where stars were hard coded (e.g. `AlbumReviews.tsx`). The `<StarsMeta />` and `<StarsMetaSpan />` components were also changed.

Because half-star ratings are allowed, the back-end has been updated to take `rating` and related members in as `double` instead of `int`.  Included in this PR is a migration script to migrate the MSSQL database from `int` to `double`.

This change was tested locally and verified to work.

### DIFF
--- a/VocaDb.Migrations/Migrations.cs
+++ b/VocaDb.Migrations/Migrations.cs
@@ -4,6 +4,26 @@ using FluentMigrator;
 
 namespace VocaDb.Migrations;
 
+[Migration(2026_05_14_1633)]
+public class ChangeRatingColumnsToDecimal : Migration
+{
+	public override void Up()
+	{
+		// Per-user rating: allow one decimal place (e.g. 0.5)
+		Alter.Column("Rating").OnTable(TableNames.AlbumsForUsers).AsDecimal(3, 1).NotNullable().WithDefaultValue(0.0m);
+
+		// Album aggregate total: allow one decimal place and larger range for totals
+		Alter.Column("RatingTotal").OnTable(TableNames.Albums).AsDecimal(9, 1).NotNullable().WithDefaultValue(0.0m);
+	}
+
+	public override void Down()
+	{
+		// Revert to previous integer columns
+		Alter.Column("Rating").OnTable(TableNames.AlbumsForUsers).AsInt32().NotNullable().WithDefaultValue(0);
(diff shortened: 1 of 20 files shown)


FinnRG (changes requested):
Hi, thank you for your contributions. A few comments on the changes:

1. The changeset is too large and contains unrelated changes. AFAICT you formatted the files using a different formatter. Please run `npm run format` or check your setup
2. I don't like pulling in an entirely different framework for a single UI component. According to this fiddle: https://jsfiddle.net/jqwidgets/6bA3D/, jqWidgets might support a precision parameter for half star reviews. Could you check that out?
3. Not a strict requirement, but it would be great if you could squash/rewrite your commit history to make it cleaner and reviews easier. It currently contains a lot of fixups that could be merged to the previous commits.

tonyhieu:
@FinnRG I addressed your changes:
1. I ran `npm run format` to get rid of irrelevant changes
2. Discussed Later
3. Commits squashed into two commits

I faced a lot of difficulty trying to get `JqxRating` to work. I agree that bringing in a new library for one component is overkill. Some of the efforts I tried with `JqxRating` (with the setting `precision={0.5}`) were:
- Changing `.cshtml` files to support fractional stars
- Adding an always-changing key to the widget in order to get it to reload on hide/show
- Importing `jQuery` in `index.tsx`
- Bumping version

In the end, none of these attempts worked, and I'm not entirely sure why. I suspect it has something to do with the way the DOM is rendered, as `jQuery` isn't built on top of React like `material-ui` and is likely facing a conflict somewhere in the code. My solution was to create a new, minimal `RatingPicker` component with the desired functionality. I think that it's the cleanest solution possible given that `JqxRating` does not want to cooperate, and `RatingPicker` can also be potentially reused elsewhere.

